PR #12, opened 2018-02-28
Drop CONDA_BUILD_SYSROOT

jakirkham:
Closes https://github.com/conda-forge/conda-smithy/issues/678

As determining the macOS SDK location happens at runtime as part of `conda-forge-ci-setup` and sets `CONDA_BUILD_SYSROOT` in the process, drop `CONDA_BUILD_SYSROOT` from `conda_build_config.yaml`. After all this setting isn't correct for Travis CI anyways.

xref: https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/3


jakirkham:
@msarahan and @isuruf, could you please double check this? Basically we want to pick up the environment variable instead. I don't know if we have to set that explicitly in here or if it happens automatically.

Issue #8, opened 2018-02-27
Add gsl


jakirkham:
Currently `gsl` is unpinned. Though according to its [ABI/API report]( https://abi-laboratory.pro/tracker/timeline/gsl/ ), it shouldn't be. We should add it to the list with some reasonable pinning to start.

cc @conda-forge/gsl

jakirkham:
Addressing in PR ( https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/22 ).

PR #22, opened 2018-03-03
Pin gsl to 2.2

jakirkham:
Fixes https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/8

The current version is 2.2.1 and both 2.2 and 2.2.1 seem compatible except for the latter adds more symbols. It seems to break API/ABI on minor versions though. Hence the 2.2 pinning instead of just 2.

cc @conda-forge/gsl 
cc @isuruf and @dougalsutherland (for an extra pair of eyes)

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -71,6 +71,8 @@ fontconfig:
   - 2.12
 freetype:
   - 2.8.1              # [not ppc64le]
+gsl:
+  - 2.2
 gstreamer:
   - 1.12
 gst_plugins_base:


dougalsutherland:
2.3 and 2.4 have been out for a while. Any reason not to update before doing this?

dougalsutherland:
Well, the reason not to is that the enormous patch used to build on Windows doesn't apply to the new version, and the source of the patch hasn't updated it. Opening an issue, might investigate later. 2.2 is good for now I guess.

jakirkham:
So I agree with the sentiment of getting `gsl` update. That said, we are making progress on `conda-build` 3, which this will be needed for. Given this, am going to put this in. However I'd encourage you to submit an update PR once newer `gsl` versions are ready.

PR #32, opened 2018-04-03
Pin sqlite to 3.22

jakirkham:
Bumps the `sqlite` pinning to 3.22 from 3.20. Also should help resolve issue ( https://github.com/conda-forge/sqlite-feedstock/issues/25 ).

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -199,7 +199,7 @@ snappy:
 sox:
   - 14.4.2
 sqlite:
-  - 3.20
+  - 3.22
 sundials:
   - 2.7
 tk:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "2018.03.18" %}
+{% set version = "2018.04.03" %}
 
 package:
   name: conda-forge-pinning


jakirkham:
Thoughts @conda-forge/core?

jjhelmus:
sqlite has been very good about keeping ABI backwards compatibility between releases according to [ABI Tracker](https://abi-laboratory.pro/tracker/timeline/sqlite/).  Would it make sense to relax this pinning to `sqlite 3.*`?  

Since symbols are occasionally added an even better pinning would be >=build_version,<major_build_version+1. Such a pinning can be done using a [pin_compatible](https://conda.io/docs/user-guide/tasks/build-packages/variants.html#pinning-expressions) Jinja function in conda build 3 or using a [run_export](https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html#run-exports) in the `sqlite` package itself.

jakirkham:
So part of the issue is 3.20 is broken and may be hard to fix.

xref: https://github.com/conda-forge/sqlite-feedstock/issues/25

dougalsutherland:
Our build of 3.20 is actually 3.19, but it’s ABI-compatible with 3.20 through 3.23 per the ABI tracker anyway, right?

It seems like build with 3.x, run with >=3.x should be fine for any value of x.

jakirkham:
Fair enough. Replacing with PR ( https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/33 ).

PR #33, opened 2018-04-09
Pin sqlite to 3

jakirkham:
Relaxes the `sqlite` pinning to 3 as `sqlite` seems to preserve backwards compatibility in major versions.

xref: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/32

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "2018.03.18" %}
+{% set version = "2018.04.03" %}
 
 package:
   name: conda-forge-pinning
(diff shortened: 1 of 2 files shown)


recipe/meta.yaml
```diff
@@ -1,4 +1,4 @@
-{% set version = "2018.03.18" %}
+{% set version = "2018.04.03" %}
```
msarahan: perhaps it is best to leave version bumps out until some tag? It seems like this will be inaccurate, because it really should be when the PR gets merged, not proposed, and it also seems like it will be highly prone to creating merge conflicts.

jakirkham: Per our discussion earlier, I'll drop the date change.

PR #37, opened 2018-04-11
Drop Win32

ocefpaf:
@isuruf can you take a look at this?

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -7,7 +7,6 @@ blas_impl:
 #    For example, we need a win-64 vs2008_win-32 package, so that we compile win-32
 #    code on win-64 miniconda.
 cross_compiler_target_platform:  # [win]
-  - win-32                     # [win]
   - win-64                     # [win]
 c_compiler:                    # [win]
   - vs2008                     # [win]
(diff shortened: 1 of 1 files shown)


ocefpaf:
> I think you need to delete Line 10 as well.

Done! Thanks!!

recipe/conda_build_config.yaml
```diff
@@ -7,7 +7,6 @@ blas_impl:
 #    For example, we need a win-64 vs2008_win-32 package, so that we compile win-32
 #    code on win-64 miniconda.
 cross_compiler_target_platform:  # [win]
-  - win-32                     # [win]
```
msarahan: this one is good to delete, but mostly not necessary.  This one is only used when creating the compiler activation packages, so it would not affect much to leave it in.

Not that you should leave it in - just explaining why it's different.

ocefpaf: Dev dooooooocs! :stuck_out_tongue_winking_eye:

recipe/conda_build_config.yaml
```diff
@@ -7,7 +7,6 @@ blas_impl:
 #    For example, we need a win-64 vs2008_win-32 package, so that we compile win-32
 #    code on win-64 miniconda.
```
jakirkham: Maybe we should drop or tweak the comment above too?

PR #38, opened 2018-04-12
hide vc behind selector for win

msarahan:
This is more important in the context of selectors finding used variables.  Without this, conda-build will consider ``vc`` used in unix recipes, because the regex is dumb regarding the "AND" condition that exists in the selectors.


msarahan:
not quite.  This plus a conda-build fix I'm putting up shortly.

PR #41, opened 2018-04-17
bump version

ocefpaf:
@conda-forge/core are we moving away from the date based release?
If so what should we version this? `1.0.0` and remove the previous releases?

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "2018.04.03" %}
+{% set version = "2018.04.16" %}
 
 package:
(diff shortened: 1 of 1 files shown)


dougalsutherland:
Naming by date is also okay I think, just name by the date when we release. It's not like there's a meaningful major/minor/patch change here so dates are just as reasonable, right?

ocefpaf:
> Naming by date is also okay I think, just name by the date when we release. It's not like there's a meaningful major/minor/patch change here so dates are just as reasonable, right?

Yep, as long as we avoid doing that every PR. I guess that was causing to many conflicts :smile:

PR #61, opened 2018-05-13
Pin glew to the minor version number

jakirkham:
As GLEW changes its SONAME every minor version, we need to pin down to the minor version. Seems patch releases don't generate much breakage (when they do occur). So just pinning to the minor version number is sufficient. Also start with the current version of GLEW in conda-forge.

ref: https://abi-laboratory.pro/tracker/timeline/glew/

Checklist


jakirkham:
cc @conda-forge/glew

Issue #69, opened 2018-05-24
pin vtk and occt


looooo:
I think it's better to pin vtk as much as possible:
`vtk: min_pin="x.x.x" max_pin="x.x.x"`

occt should be ok with this:
`occt: min_pin="x.x" max_pin="x.x"`

dougalsutherland:
Yeah, `vtk` definitely needs [full pinning](https://abi-laboratory.pro/tracker/timeline/vtk/), it removes symbols in patch releases all the time.

Does anything / is anything expected to link against occt?

looooo:
> Does anything / is anything expected to link against occt?

mainly pyocct and freecad (both not yet ready, but on the way).

looooo:
https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/70

jakirkham:
So does [occt]( https://abi-laboratory.pro/?view=timeline&l=opencascade ). Though they don't seem to have patch releases AFAICT.

PR #70, opened 2018-05-24
pin vtk and occt

looooo:
Checklist

Closes https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/69

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -172,6 +172,8 @@ pin_run_as_build:
     max_pin: x.x
   nettle:
     max_pin: x.x
+  occt:
+    max_pin: x.x
   openblas:
     max_pin: x.x.x
   openjpeg:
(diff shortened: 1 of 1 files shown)


recipe/meta.yaml
```diff
@@ -1,4 +1,4 @@
-{% set version = "2018.05.22" %}
+{% set version = "2018.05.24" %}
```
jakirkham: Do we want to do the version bump here or were there other things we need to get in first?

cc @conda-forge/core

scopatz: We can always do another release.  I don't think it is a big issue

scopatz:
Thanks for putting this in @looooo. Was occt supposed to be here?

looooo:
yes I thought it will help for packages depending on occt using the same version. Doesn't it make sense to have it pinned?

scopatz:
Ok, I just didn't see occt in the changeset

looooo:
ah you are right, something went wrong

looooo:
Is there a way to use a newer occt if this PR is accepted and a newer occt is uploaded?
I think the `max_pin: x.x` is absolutely needed for occt to not include some incompatible versions.

jakirkham:
@conda-forge-admin, please re-render.

jakirkham:
Thanks @looooo :)

PR #71, opened 2018-05-25
Bump `boost` and `boost-cpp` to `1.67.0`

jakirkham:
Checklist
* [ ] [Re-rendered]( https://conda-forge.org/docs/conda_smithy.html#how-to-re-render ) with the latest `conda-smithy`

### DIFF
--- a/.circleci/run_docker_build.sh
+++ b/.circleci/run_docker_build.sh
@@ -29,7 +29,7 @@ if [ -z "$CONFIG" ]; then
     exit 1
 fi
 
-test -d "$ARTIFACTS" || mkdir "$ARTIFACTS"
+mkdir -p "$ARTIFACTS"
 DONE_CANARY="$ARTIFACTS/conda-forge-build-done-${CONFIG}"
 rm -f "$DONE_CANARY"
 
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -216,9 +216,9 @@ pin_run_as_build:
 arpack:
   - 3.5
 boost:
-  - 1.66.0
+  - 1.67.0
 boost_cpp:
-  - 1.66.0
+  - 1.67.0
 bzip2:
   - 1
 cairo:


CJ-Wright:
Does this need a re-build of downstream libs?

jakirkham:
Boost changes to SONAME every version. Not to mention the other things it breaks each release. 😞 At least they are up front about it though.

ref: https://abi-laboratory.pro/tracker/timeline/boost/

jakirkham:
There are some serious fixes for Boost Beast and Boost Stacktrace. Also a few fixes for C++17, which will be important when using the new compilers.

jakirkham:
Thoughts @conda-forge/core?

jakirkham:
Not too surprising TBH. Boost is C++'s equivalent to the standard library. If you are writing C++ code, you are almost certainly using Boost.

Should add we are a little slow getting to this actually as Boost 1.67.0 came out back in April and we are almost in June.

jakirkham:
FWIW don't think we should need to build anything downstream of those libraries. AFAIK there hasn't been a need to do that historically. That said, what might be nice is to see how things in that list are related in topo order. People involved in those stacks probably have a good idea, but it wouldn't hurt to have that graph. Especially if we hope to automate this sort of thing in the future. 😉

xref: https://github.com/regro/cf-scripts/issues/44

dougalsutherland:
Also: make sure you're looking at a list of things with `run` deps on `boost`/`boost-cpp`, not just `build`/`host`. Lots of parts of boost are headers-only, and packages using it that way don't need a rebuild.

bdice:
I am currently trying to add a recipe to conda-forge, and boost v1.66 will not build with the conda-forge compilers due to a known issue in boost v1.66. https://svn.boost.org/trac10/ticket/13386

After this PR is merged, I will be able to fix my staged recipe. https://github.com/conda-forge/staged-recipes/pull/6130

scopatz:
This has been open long enough. Merging.

jakirkham:
Thanks all :)

PR #75, opened 2018-05-30
1.12 is not available for osx.

matham:
Checklist

This changes the pinning to use 1.14 on osx because 1.12 is not available foe osx. See https://github.com/conda-forge/staged-recipes/pull/5966#issuecomment-393048712

### DIFF
--- a/.circleci/run_docker_build.sh
+++ b/.circleci/run_docker_build.sh
@@ -29,7 +29,7 @@ if [ -z "$CONFIG" ]; then
     exit 1
 fi
 
-test -d "$ARTIFACTS" || mkdir "$ARTIFACTS"
+mkdir -p "$ARTIFACTS"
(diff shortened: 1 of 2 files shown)


matham:
Thinking about it, the next build will be for 1.14.1, not 1.14, which is currently available on anaconda. Should I maybe bump it to 1.14.1 instead in anticipation? Maybe it should be a range? I that even possible?

jakirkham:
SGTM. Thanks @matham :)

jakirkham:
FWIW they are both identical from an ABI/API standpoint. So either are fine.

Existing versions may help you remove the immediate roadblock.

ref: https://abi-laboratory.pro/tracker/timeline/gstreamer/

matham:
Ok, thanks. I guess I'll leave it at 1.14 for now then.

matham:

jakirkham:
Thanks @matham :)

PR #83, opened 2018-06-29
pin kealib

ocefpaf:
@gillins is this OK for `kealib`?

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -114,6 +114,8 @@ pin_run_as_build:
     max_pin: x.x
   jsoncpp:
     max_pin: x.x.x
+  kealib:
+    max_pin: x.x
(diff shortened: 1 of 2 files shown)


gillins:
Looks good to me :smile:

gillins:
The range of valid versions is going to be a bit out now that we are moving to C++11 mode - newly built packages are no longer going to work with older ones compiled in C++98 mode (as discovered with `kealib`). Not sure what we can do but we are going across a bit of a boundary here....

ocefpaf:
We can put a tighter pin if you think it will help. However, at the moment I'm thinking about pinning `gdal` to `kealib 1.4.7` to avoid the `hdf5 1.10.2` issues.

Once we build all packages with `hdf5 1.10.2` we can start using the new syntax and let the pin here apply.

PR #84, opened 2018-07-01
zeromq 4.2.5

minrk:
Since ABI fixes in the recipe, this *should* work with `>=4.2.5,<4.3`. <del>But I'm not sure if we want to do that right now.</del> Updated zeromq max_pin to x.x, so it should get the right pinning on build.

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -217,7 +217,7 @@ pin_run_as_build:
   xz:
     max_pin: x.x
   zeromq:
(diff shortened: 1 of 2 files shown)


ocefpaf:
@minrk if you want this to be in effect "right now" please bump the version (the date in the `meta.yaml`) and I'll merge it. If not, we can merge and wait for more changes before a new release.

minrk:
@ocefpaf updated date and fixed max_pin for zeromq now that the ABI issue is fixed in 4.2.5

PR #91, opened 2018-07-14
Add OpenTURNS pinning

jschueller:
Checklist
* [ ] [Re-rendered]( https://conda-forge.org/docs/conda_smithy.html#how-to-re-render ) with the latest `conda-smithy`

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -373,6 +373,8 @@ openjpeg:
   - 2.3
 openssl:
   - 1.0.2
+openturns:
+  - 1.11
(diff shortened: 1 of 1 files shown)


jschueller:
We dont have proper versioning yet, I guess each release is incompatible with the other.
Except for otpod, batman, all should be pinned to 1.11 exactly.

dougalsutherland:
Went ahead and added a `pin_run_as_build` entry too then, so that downstream packages will stick to the minor version they were built with. You could also/alternatively add a `run_exports` entry to the recipe.

PR #93, opened 2018-07-14
specify mpi provider matrix

minrk:
so recipes can have just `{{ mpi }}` and get both openmpi and mpich without needing their own conda_build_config.yaml. It seems like this is a reasonable thing, since most packages seem to be adding mpi variants? If not, feel free to close and packages can add their mpi variants in their own conda_build_config.yaml as they are doing now.

Checklist


ocefpaf:
Thanks @minrk!

PR #94, opened 2018-07-15
Python 3.7

ocefpaf:
I want to get #93 in.

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -19,6 +19,7 @@ c_compiler:
   - vs2008                     # [win]
   - vs2015                     # [win]
   - vs2015                     # [win]
+  - vs2015                     # [win]
 cxx_compiler:
   - toolchain_cxx              # [linux or osx]
   # new compilers
@@ -30,7 +31,7 @@ cxx_compiler:
   - vs2008                     # [win]
   - vs2015                     # [win]
   - vs2015                     # [win]
-
+  - vs2015                     # [win]
 fortran_compiler:              # [unix or win64]
   - toolchain_fort             # [unix and not ppc64le]
   - gfortran                   # [(linux or osx)                      and (environ.get('CF_COMPILER_STACK') == 'comp7')]
@@ -447,6 +448,7 @@ proj4:
   - 4.9.3
 python:
   - 2.7
+  - 3.5
   - 3.6
   - 3.7
 pyqt:
@@ -469,6 +471,7 @@ tk:
   - 8.6                # [not ppc64le]
 vc:                    # [win]
   - 9                  # [win]
+  - 14                 # [win]
   - 14                 # [win]
   - 14                 # [win]
 vlfeat:
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -19,7 +19,7 @@ c_compiler:
   - vs2008                     # [win]
   - vs2015                     # [win]
   - vs2015                     # [win]
-  - vs2015                     # [win]
+  - vs2015                     # [win and (environ.get('CF_MAX_PY_VER', '') >= '37')]
 cxx_compiler:
   - toolchain_cxx              # [linux or osx]
   # new compilers
@@ -31,7 +31,7 @@ cxx_compiler:
   - vs2008                     # [win]
   - vs2015                     # [win]
   - vs2015                     # [win]
-  - vs2015                     # [win]
+  - vs2015                     # [win and (environ.get('CF_MAX_PY_VER', '') >= '37')]
 fortran_compiler:              # [unix or win64]
   - toolchain_fort             # [unix and not ppc64le]
   - gfortran                   # [(linux or osx)                      and (environ.get('CF_COMPILER_STACK') == 'comp7')]
@@ -450,7 +450,7 @@ python:
   - 2.7
   - 3.5
   - 3.6
-  - 3.7
+  - 3.7  # [(environ.get('CF_MAX_PY_VER', '') >= '37')]
 pyqt:
   - 5.6
 qt:
@@ -473,7 +473,7 @@ vc:                    # [win]
   - 9                  # [win]
   - 14                 # [win]
   - 14                 # [win]
-  - 14                 # [win]
+  - 14                 # [win and (environ.get('CF_MAX_PY_VER', '') >= '37')]
 vlfeat:
   - 0.9.20
 vtk:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "2018.08.05" %}
+{% set version = "2018.08.14" %}
 
 package:
   name: conda-forge-pinning


recipe/conda_build_config.yaml
```diff
@@ -390,6 +390,7 @@ python:
   - 2.7
   - 3.5
   - 3.6
+  - 3.7
```
isuruf: You have to change the compilers for Windows too. Let's discuss this a bit before merging.

ocefpaf: Not sure what you mean? Are the new Python 3.7 package built with another compiler?

ocefpaf: BTW, we are sprinting here are SciPy and there are tons of people waiting on these features that are being blocked by "let's wait...

isuruf: > Not sure what you mean? Are the new Python 3.7 package built with another compiler?

No, it's still vs2015, but you have to tell conda-build that.

CJ-Wright:
What do we need to do in the way of rebuilds once this is in?

ocefpaf:
I guess we need to repeat the `vs2015` lines, is that right @msarahan?

recipe/conda_build_config.yaml
isuruf: > BTW, we are sprinting here are SciPy and there are tons of people waiting on these features that are being blocked by "let's wait..

My concern was that this would suddenly start creating python 3.7 builds for every package and it would pull dependencies from defaults leading to incompatibilities. I like the idea of the bot making PRs.

I don't know what you mean by being blocked. If you want 3.7 builds, it's as simple as adding a `conda_build_config.yaml` to the recipe folder and rerendering.

msarahan:
They're zipped, and all zipped key fields must be the same length.  Yes, you must add vs2015 to c_compiler and cxx_compiler, because you are adding a python entry, and you must make the associated zipped fields match that.

recipe/conda_build_config.yaml
ocefpaf: Yes, the new compiler migration is blocked on a "forever wait" plead. We need to get the ball rolling there.

recipe/conda_build_config.yaml
```diff
@@ -390,6 +390,7 @@ python:
   - 2.7
   - 3.5
```
jakirkham: Could we drop Python 3.5?

ocefpaf: Not now.

isuruf:
@CJ-Wright, I don't think you have to wait for this. what the bot can do is start from the dependencies with only python dependencies and start issuing PRs. You just have to add the following to `recipe/conda_build_config.yaml` and not commit that file.

```
python
- 2.7
- 3.5
- 3.6
- 3.7
c_compiler:  #[win]
- vs2008  #[win]
- vs2015  #[win]
- vs2015  #[win]
- vs2015  #[win]
cxx_compiler:  #[win]
- vs2008  #[win]
- vs2015  #[win]
- vs2015  #[win]
- vs2015  #[win]
```

isuruf:
Also, should we send `--skip-existing` to conda-build if it's a master branch build? @msarahan, is there an option to skip if it exists in `conda-forge/label/main`?

msarahan:
is conda-forge/label/main the same as conda-forge?  If so you don't need to do anything more than pass --skip-existing.  Conda-build considers all the channels you have currently configured when determining if something exists.

CJ-Wright:
@isuruf I'm confused why are we adding a file but not committing it? Won't that do nothing? (The bot doesn't run conda-build itself). Do we care if they are run or host dependencies on python?

Also
Attn: @justcalamari

isuruf:
> is conda-forge/label/main the same as conda-forge?

yes

> If so you don't need to do anything more than pass --skip-existing. Conda-build considers all the channels you have currently configured when determining if something exists.

Will it skip if the build is in defaults, but not conda-forge?

isuruf:
> @isuruf I'm confused why are we adding a file but not committing it? Won't that do nothing? (The bot doesn't run conda-build itself). Do we care if they are run or host dependencies on python?

It will make a change, because conda-smithy will create a config each for all python versions including 3.7. I suggested not committing the `conda_build_config.yaml` file, because we don't want recipes limited to those python versions for future python versions.

isuruf:
> I suggested not committing the conda_build_config.yaml file, because we don't want recipes limited to those python versions for future python versions.

And also for not limiting defaults or other people using our recipes.

msarahan:
It will skip if and only if some package matches the exact package name.  Any difference in hash (due to different used variables, for example) will end up getting built.

recipe/conda_build_config.yaml
jakirkham: Why not?

recipe/conda_build_config.yaml
mariusvniekerk: We could stage this (and probably the dual compiler) somewhere special and have the bot do both of these in graph order together?

jakirkham:
To recap my understanding of our discussion in the meeting earlier, we have two migrations that are needed.

1. Rebuild with new compilers
2. Migrate to Python 3.7

It sounds like we will prioritize 1. However if there are packages that need both (e.g. `numpy`), we will try to handle them first.

recipe/meta.yaml
```diff
@@ -1,4 +1,4 @@
-{% set version = "2018.07.01" %}
+{% set version = "2018.07.15" %}
```
scopatz: This version number is out of date

CJ-Wright:
I think we may also be able to do symbiotic migrations if that would be desirable. So we can run a compiler migration and move to py37 in one PR. The upside is that we reduce the amount of CI usage, but then one may need to debug both at once.

isuruf:
Packages without compilers that needs python 3.7 will not be that large I think because most of them will be noarch.

jakirkham:
Getting some numbers would help :)

recipe/conda_build_config.yaml
jakirkham: Given the other builds needed for the new compilers, I don't think it is reasonable to hang on to Python 3.5 when adding Python 3.7.

CJ-Wright:
We could also start migrating those py37 things which have no compiler dep. in a separate migrator. This would mean that we'd have both the compiler+py37 and py37 only running at the same time. Would we need to rebuild the py37 only packages after the compiler migration?

isuruf:
Are we going to support numpy 1.9 for python 3.7? If not, we should fix numpy entry here

scopatz:
That is probably up to the numpy maintainers, but I don't see why we wouldn't. We support 1.9 for all other Pythons.

jakirkham:
Are we going to drop Python 3.5?

scopatz:
@jakirkham - it would be nice to drop v3.5, but that could be a separate PR.

ocefpaf:
> it would be nice to drop v3.5, but that could be a separate PR.

We cannot drop PY35 immediately. Some people at SciPy brought to us that many are depending on it to build stuff on CIs and, as the package are no longer updated, the CIs testing may break. We need to announce it and give 2-3 month time period so people can adapt their systems.

jakirkham:
So this is increasing the load on the CIs while we are also planning to double up the builds for the new compilers. Given the messages we have been getting from CircleCI, this is even more important to consider carefully. Would ask that people consider what our priorities are in terms of are limited resources.

ocefpaf:
> So this is increasing the load on the CIs while we are also planning to double up the builds for the new compilers. Given the messages we have been getting from CircleCI, this is even more important to consider carefully. Would ask that people consider what our priorities are in terms of our limited resources.

We are all aware of that and I'm the first one to request to drop stuff, like for Python 3.4 and Windows 32, and, although I appreciate that now you are on the other side and understand my reasons to ask to drop those, I now ask you to understand that 2-3 months is not bad for our CIs at all.

isuruf:
Can we use `--skip-existing` to avoid building the package for older versions if it exists?

scopatz:
Yeah, I think that frankly we are just in a difficult spot that none of asked for.  No one in their right mind wants a compiler ABI incompatibility at the same time as a Python minor release when the CIs are threatening to turn off the compute.

My hope is that we can get some support (from NumFOCUS) to make all of these problems a little more tenable going forward. But I also feel like we can't let one problem block others, or we won't make any progress on any of them. We are a large enough organization that we should be able to walk and chew gum at the same time. 

Also, I think that @CJ-Wright @justcalamari @mariusvniekerk and everyone who has been helping out on the bot, especially with regards to the no-arch migrator, has helped buy us some much needed CI time lattitude.

jakirkham:
> Can we use `--skip-existing` to avoid building the package for older versions if it exists?

Sounds like issue ( https://github.com/conda-forge/conda-smithy/issues/234 ).

jakirkham:
To be clear, I didn't say I was "blocking" this or that I have a personal opinion on it. Only that I wanted us to be aware of the issues that this presents.

Does [our original plan]( https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/94#issuecomment-405726832 ) still hold or are there changes planned to that as well?

scopatz:
> To be clear, I didn't say I was "blocking" this or that I have a personal opinion on it. Only that I wanted us to be aware of the issues that this presents.

Ahh gotcha :smile: 

> Does our original plan still hold or are there changes planned to that as well?

I think that plan more or less still holds.  But I'll add some thoughts to it. 

* I believe that the compiler migratation (https://github.com/conda-forge/conda-smithy/pull/836) is stalled for reasons I haven't been following carefully enough to understand.
* I don't believe that the compuiler migration should block other migrations, especially if there is a clearly demonstrated need.
* I don't know of anyone claiming that they **need** the Python 3.7 at this point. (Perhaps @asmeurer asked for something something somewhere?)

So if we or our users start really needing Python 3.7, I don't think we should prevent that migration just because we haven't done the compilers yet.

Also, I am in favor of declaring Oct 1. as the sunset date for Python 3.5 right now. That would give downstream packages more than ehough time to move their CIs.

asmeurer:
The sooner conda-forge gets 3.7 packages the better. We don't actually have CI testing for it yet because of this.

Also, I hope you guys came to your senses and aren't actually going to delete the 3.5 packages from the repos.

mariusvniekerk:
@asmeurer , as far as I recall it was more along the lines of using some of the repodata.json rewrite ideas to provide a conda-forge channel with some of the older pieces excluded, without touching the real channel

msarahan:
Yes, Phil is adamant that packages never be removed. This is a repodata
optimization thing. Since conda forge is so big, it is making the solver
much slower.

asmeurer:
Is the idea that the default `conda-forge` channel would still have the 3.5 packages or not?

pelson:
> Is the idea that the default conda-forge channel would still have the 3.5 packages or not?

I don't think you will get an answer to this question here, and I don't think it is the right place to discuss it.

My opinion is that py35 packages should eventually be removed from the primary channel primarily for size of repodata payload, but also to a lesser extent for reduced conda resolution workload reasons. I don't *ever* support conda-forge binaries being *deleted*, so you should always be able to download binaries that you've already resolved/downloaded, but I do support the index being pruned of packages that are obsolete/compromised/incorrect so that conda never resolves these packages when constructing a new environment. I also have no problem whatsoever with us having sub-channels that continue to include obsolete binaries in their repodata (e.g. ``conda-forge/label/py35-legacy``) for legacy reasons - to be clear though, there is no chance I'd support us offering to continue to build new stuff for these channels.

I'm sure there are other opinions on this, so don't take what I say as gospel. I suggest that this is **not** the place to discuss this idea - happy to be involved in a discussion, but please ping me elsewhere.

mariusvniekerk:
Looks good to me in the current state.

asmeurer:
So where is the correct place to discuss it?

msarahan:
@ocefpaf zip_keys does suck.  If you have suggestions for a better way, it is definitely an area ripe for improvement.

ocefpaf:
> @ocefpaf zip_keys does suck. If you have suggestions for a better way

Who said that :eyes: :smile: 

>  it is definitely an area ripe for improvement.

I opened #111 to add tests here so we don't forget to add them. Not sure if we can do better than that.

msarahan:
Too bad you can't force-push to my email account!

jakirkham:
> So where is the correct place to discuss it?

Maybe add a new issue on the webpage repo?

jakirkham:
Sorry I missed this. What is `CF_MAX_PY_VER`? Where is this being set? What is it doing?

mariusvniekerk:
@jakirkham the idea is to gate the py37 behind an environment variable that we can control from conda-smithy.  This allows us to move to 3.7 when certain feedstocks are ready to move once their dependencies have been migrated

ocefpaf:
This one went from adding Python 3.7 to re-add Python 3.5 :smile:

@mariusvniekerk are we waiting on `conda-smithy` to merge this one?

There are some important pinning updates done by @bgruening that we need to release.

PR #98, opened 2018-07-18
add x264 pin on specific sonumber

carlodri:
Checklist
* [x] Used a fork of the feedstock to propose changes
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/conda_smithy.html#how-to-re-render ) with the latest `conda-smithy`

Fixes #51 

CC @conda-forge/x264

PR #101, opened 2018-07-19
x264 pin (wrongly?) interpreted as strict

carlodri:
Checklist

After the release of #98, recipe at https://github.com/conda-forge/ffmpeg-feedstock/pull/51 seems to pick up the `1!152` pin as a strict pin, rather than the expected fuzzy pin like `1!152.*`.

Relevant CI: https://travis-ci.org/conda-forge/ffmpeg-feedstock/jobs/405819733#L378

This PR is an attempt (probably rough) to fix this bringing back the wildcard in the pin (as suggested by @sdvillal [here](https://github.com/conda-forge/ffmpeg-feedstock/pull/51#issuecomment-406357562). Better solutions and enlightenments on the matter are very welcome. 

CC @dougalsutherland @conda-forge/x264


carlodri:
@scopatz thanks to you!

Sadly, the modification seems to have had no effect as the linter bot says:

> I rerendered the feedstock and it seems to be already up-to-date.

carlodri:
I'm not sure but this could be pointing to a bug/inconsistent behavior _somewhere_ in the conda toolchain handling version parsing (probably [epochs](https://www.python.org/dev/peps/pep-0440/#id28)?).  It appears that `1!152` is not parsed in the same "fuzzy" way as, _e.g._ `1.4`. 

IIUC, `1.4` allows for `1.4.1`, `1.4.2` and so on. Similarly, `1!152` should allow for `1!152.20180712`, whereas it does not (see CI linked in the PR description), and strictly pins to `1!152`.

CC @conda-forge/conda-forge-pinning (maybe someone else?)

183amir:
I guess we can turn this to an issue on conda-build.

PR #103, opened 2018-07-24
pin jasper

ocefpaf:
@msarahan what is the approach you are using on defaults for jasper?
We are not upgrading our version just yet...

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -115,6 +115,8 @@ pin_run_as_build:
     max_pin: x
   isl:
     max_pin: x.x
+  jasper:
+    max_pin: x.x.x
   jpeg:
     max_pin: x
   json-c:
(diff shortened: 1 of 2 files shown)


dougalsutherland:
It seems like it's very stable: https://abi-laboratory.pro/?view=timeline&l=jasper. Why is the super-tight pin needed?

msarahan:
We don't currently pin it in conda_build_config.yaml.  It hasn't been updated since last September, why the attention now?  http://www.ece.uvic.ca/~frodo/jasper/#download

msarahan:
ah, I see.  You still use the old version.

ocefpaf:
> ah, I see. You still use the old version.

Yep. The new version has new dependencies that we are not packaging yet.

@dougalsutherland can you take a second look?

ocefpaf:
> Assuming we're not doing any further 1.x releases, this is definitely fine. 

Nope. I'm working on 2.x

> Presumably fine even if not but idk.

Probably.

PR #114, opened 2018-08-30
re-pin libnetcdf

ocefpaf:
See https://github.com/Unidata/netcdf4-python/issues/836

I'll merge this tomorrow, pull all the `libnetcdf >4.4.x`, and start re-building stuff :unamused: 

Ping @dopplershift

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -376,7 +376,7 @@ libkml:
 libmatio:
   - 1.5
 libnetcdf:
-  - 4.6
+  - 4.4
 libpcap:
   - 1.8
 libpng:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "2018.08.28" %}
+{% set version = "2018.08.30" %}
 
 package:
   name: conda-forge-pinning


recipe/conda_build_config.yaml
```diff
@@ -376,7 +376,7 @@ libkml:
 libmatio:
   - 1.5
 libnetcdf:
-  - 4.6
+  - 4.4
```
efiring: Did you mean 4.5?

ocefpaf: Nope. I decided to go with 4.4 b/c of a mismatch on Windows Python 2.7, there is no 4.5 there due to modern c++. We patched 4.6 to make that happen but it would be a lot of work to backport those patches to 4.5.

The main issue will be installing `nco` in the same env as `netcdf4` b/c `nco` does require `4.5`. So for now people that need `nco`'s latest features will have to create separate envs.

ocefpaf: Reading https://github.com/Unidata/netcdf4-python/issues/836 looks like it is something we can "fix" by setting a higher timeout. If that is possible I'd rather build a new `4.6.1` than re-pinning and re-build all the stack.

@dopplershift what do you think?

ocefpaf:
Let's try https://github.com/conda-forge/libnetcdf-feedstock/pull/53 first. I don't really want to rebuild stuff.

PR #21, opened 2018-03-03
Bump glpk to 4.65

jakirkham:
Fixes https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/13

cc @dougalsutherland

PR #120, opened 2018-09-11
Bump glib pin to 2.56, matching defaults

pkgw:
As per #21.


ocefpaf:
Let's bump the package version too to get that in. (The other two pending PRs are kind of stale and we should not wait from them.)

pkgw:
Like so?

PR #122, opened 2018-09-20
Pin hdf5 to 1.10.3

tschoonj:


recipe/conda_build_config.yaml
```diff
@@ -341,7 +341,7 @@ harfbuzz:
 hdf4:
   - 4.2
 hdf5:
-  - 1.10.2
+  - 1.10.3
```
ocefpaf: We don't pin software here unless it **must** diverge from the global pinning.

You should send a PR to https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/master/recipe/conda_build_config.yaml#L344 instead so all packages will get re-pinned.

ocefpaf: Jeez I clicked on your hdf5 PR and commented here. Sorry. Disregard what I wrote above. This is correct!

ocefpaf:
Please bump the build number to get a new package.

scopatz:
Thanks @tschoonj and @ocefpaf!

tschoonj:
Cool: so do I open a PR in h5py-feedstock to ensure a new build gets made against hdf5 1.10.3 or does this happen automagically?

scopatz:
Yeah, you can open an rerender PR if you want to ensure this version of hdf5 is used.  Otherwise it will happen automatically whenever h5py gets rebuilt next.

PR #125, opened 2018-09-26
Drop HDF5 pin back to 1.10.2; bump version to 2018.09.26

### LINKED ISSUE #124: Drop hdf5 pin back to 1.10.2?
As per [h5py-feedstock #35](https://github.com/conda-forge/h5py-feedstock/pull/35). It looks like the upgrade to HDF5 1.10.3 may have been a bit premature, as the h5py test suite fails when built against it.

CC @jakirkham .

pkgw:
HDF5 1.10.3 [causes problems for
h5py](https://github.com/conda-forge/h5py-feedstock/pull/33#issuecomment-423360989)
so we don't want to start using later versions until these problems are ironed
out.

Closes #124.

Checklist


ocefpaf:
Maybe we should pin h5py's meta.yaml instead of reverting this. As far as I
know 1.10.3 has some security updates and most of the packages are OK with
it.

scopatz:
Yeah, that would be better, ultimately, if h5py cannot be patched and does not make a timely release. Are they aware of the issue?

jakirkham:
cc @conda-forge/h5py

jakirkham:
If `h5py` cannot be installed with other parts of the stack due to `hdf5` version mismatch, that’s problematic as well.

scopatz:
Sure, but if h5py pins an hdf5 version, won't most other parts of the stack be OK with using that hdf5 version?

scopatz:
Also, it is kind of weird to me that they did something that was API breaking enough to cause h5py to fail on a micro version number bump.

jakirkham:
The issue is known to `h5py` developers. ( https://github.com/h5py/h5py/issues/1088 ) Though there is no known fix AFAIAA.

I’m not sure. HDF5 needs to be pinned at the patch version. So packages would need to be rebuilt with this exact version. As two people have sent PRs within a day of each other requesting `h5py` update it’s `hdf5` version, it suggests there’s more to the story here.

While there are some invalid pointer dereferencing issues and some zero division issues, the CVEs do not seem so severe that we need to push the stack to the newest version before we are ready.

tacaswell:
The change is that different exceptions are coming out of the guts.  

If CF wants to push forward I suggest applying a patch that removes / known-fails those two tests.

PR #130, opened 2018-10-08
Pin nlopt

jschueller:
hello,
nlopt's commit
https://github.com/stevengj/nlopt/commit/61fe5161861c7658c70563fbf3abf456fb3ce980
adds get_errmsg (after 2.4.2), and it's used in the public cxx interface, so libs linked against nlopt 2.5.0 cannot be used with 2.4, I experienced it with openturns packages today (https://github.com/conda-forge/openturns-feedstock/pull/61)
As there are other c++ packages here (https://github.com/search?l=YAML&q=org%3Aconda-forge+nlopt&type=Code) that depend on nlopt I propose to move the pin here.

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -227,6 +227,8 @@ pin_run_as_build:
     max_pin: x.x
   nettle:
     max_pin: x.x
+  nlopt:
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks @jschueller!

PR #133, opened 2018-10-12
pin libpq

sdvillal:
Checklist
* [X] Used a fork of the feedstock to propose changes

Would it be worth to pin libpq here, so downstream dependencies like psychopg2 get consistent on the stack?

Attending to the postgres ABI lab (https://abi-laboratory.pro/index.php?view=timeline&l=postgresql)
I think x.x is the right pin.

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -191,6 +191,8 @@ pin_run_as_build:
     max_pin: x.x
   libpng:
     max_pin: x.x
+  libpq:
+    max_pin: x.x
   libprotobuf:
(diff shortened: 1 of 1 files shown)


sdvillal:
On a second thought, likely this is a bad idea. Migrating databases just to keep consistent with the stack is not something we should enforce our users to do. Instead, we should make sure we support a few major versions of packages like postgres (9, 10), we export proper pins with each version (which in postgres case would mean to backport the good work done on splitting the recipe to a branch "version 9") and perhaps declare these versions as variants here.

Closing

PR #137, opened 2018-10-16
update libwebp pin

ocefpaf:
The bot built `libwebp 1.0.0` so instead of keeping an old version in a branch let's bump the pin.


scopatz:
LGTM!

PR #141, opened 2018-10-31
Bump gdal

ocefpaf:
Now that `fiona 1.8` was released we can pin the stack to `gdal 2.3`.

Pinging @msarahan and @jjhelmus who may want to do the same in `defaults`.


scopatz:
LGTM

jjhelmus:
Thanks for the heads up @ocefpaf, I'll look into updating the pin in `defaults`.

PR #150, opened 2018-11-16
pin kealib and poppler

ocefpaf:
We need this to get https://github.com/conda-forge/gdal-feedstock/pull/237 going on `gcc7`.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "2018.11.13" %}
+{% set version = "2018.11.19" %}
 
 package:
   name: conda-forge-pinning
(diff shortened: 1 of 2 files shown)


recipe/meta.yaml
```diff
@@ -1,4 +1,4 @@
-{% set version = "2018.11.13" %}
+{% set version = "2018.11.19" %}
```
ocefpaf: Let's wait until Monday b/c I would like to hear what @gillins has to say about the `kealib` pin.

bgruening: @gillins confirmed in https://github.com/conda-forge/gdal-feedstock/pull/237 this is ok. So we could get this in.

ocefpaf:
Now that we have a `cartopy` that works with `proj.4 5.x` let me add that change here to before we merge.

PR #166, opened 2018-12-28
Specify that dual compiler is only for linux64, not linux-aarch64

hmaarrfk:
I would like to make this compatible with aarch64.

A few things don't make sense for aarch64, specifically, the dual compiler stream.

I haven't added any `[aarch64]` in this pr, but rather made `linux` pretty specific to `linux64`, which is all that conda-forge compiles for today.

Checklist

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -9,7 +9,7 @@ blas_impl:
 cross_compiler_target_platform:  # [win]
   - win-64                     # [win]
 c_compiler:
-  - toolchain_c                # [linux or osx]
+  - toolchain_c                # [linux64 or osx]
   # new compilers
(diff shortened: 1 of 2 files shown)


hmaarrfk:
No, I think this is mostly to avoid doing a "dual compiler" setup again for aarch64.

I think @jjhelmus will be helping me get a real compiler up and running soon, and we will be able to see what gets built, and what doesn't.

I think most selectors in the recipes should be fine (I hope). We might might need to decide if we want the graphical stack for aarch64. but we can document those changes as they come.

mariusvniekerk:
@hmaarrfk i've effectively added this in another pr. #173

hmaarrfk:
ok.

PR #173, opened 2019-01-12
Add builds support for aarch64

mariusvniekerk:
Tested with

* https://github.com/conda-forge/zlib-feedstock/pull/26

Example builds

*  https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=5267&view=logs


mariusvniekerk:
well it's harmless to merge in.  The big thing that is important to note is the change for using `linux64` instead of `linux` as we need to be a bit more platform-arch compatible

PR #176, opened 2019-01-16
Update harfbuzz pin to 2.x series and bump version to 2019.01.16.

pkgw:
ABI Laboratory shows that the 2.x series has been backward-compatible since the 2.0 release. We need to start using the newer harfbuzz to start using the newer cairo (namely the 1.16 series) in places like gtk-2 and librsvg.

CC @croth1 and conda-forge/gtk2-feedstock#8: I ran into another issue where this old pin was holding things back, so I went ahead and put together the PR.


pkgw:
CI failed due to something spurious.

jakirkham:
Toggling for CI.

PR ( https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/178 ) should have fixed it. Though we may need to explicitly merge `master` here if toggling did not work.

pkgw:
CI indeed fixed.

PR #179, opened 2019-01-18
Pin gdal

ocefpaf:
Ping @hobu and @kmuehlbauer


ocefpaf:
Should fix https://github.com/conda-forge/basemap-feedstock/issues/43, https://github.com/conda-forge/rasterio-feedstock/issues/98, and https://github.com/conda-forge/pdal-feedstock/pull/54 after a re-render.

PR #182, opened 2019-01-29
Advances gstreamer pin to 1.14.x

croth1:
Checklist

My first attempt to advance pins. Can someone verify that this is done correctly? 
Version 1.14.0 is now pinned, 1.14.x can be used during runtime, due to this code https://github.com/conda-forge/gstreamer-feedstock/blob/master/recipe/meta.yaml#L26

I removed the max_pin, because it seems to be transferred to gstreamer already.

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -141,8 +141,6 @@ pin_run_as_build:
     max_pin: x
   gsl:
     max_pin: x.x
-  gst-plugins-base:
-    max_pin: x.x
   gdal:
     max_pin: x.x
   geos:
@@ -330,8 +328,7 @@ gsl:
 gstreamer:
   - 1.14.4
 gst_plugins_base:
-  - 1.12  # [linux]
-  - 1.14.0  # [osx]
+  - 1.14.4
 gdal:
   - 2.4
 geos:
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -141,8 +141,6 @@ pin_run_as_build:
     max_pin: x
   gsl:
     max_pin: x.x
-  gstreamer:
-    max_pin: x.x
   gst-plugins-base:
     max_pin: x.x
   gdal:
@@ -154,8 +152,6 @@ pin_run_as_build:
   giflib:
     max_pin: x.x
   glew:
-    max_pin: x.x
-  glib:
     max_pin: x.x
   glpk:
     max_pin: x.x
@@ -332,8 +328,7 @@ gf2x:
 gsl:
   - 2.4
 gstreamer:
-  - 1.12  # [linux]
-  - 1.14.0  # [osx]
+  - 1.14.4
 gst_plugins_base:
   - 1.12  # [linux]
   - 1.14.0  # [osx]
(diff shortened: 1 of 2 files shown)


recipe/conda_build_config.yaml
```diff
@@ -141,8 +141,6 @@ pin_run_as_build:
     max_pin: x
   gsl:
     max_pin: x.x
-  gstreamer:
-    max_pin: x.x
```
scopatz: I think this should remain.

recipe/conda_build_config.yaml
```diff
@@ -332,8 +330,7 @@ gf2x:
 gsl:
   - 2.4
 gstreamer:
-  - 1.12  # [linux]
-  - 1.14.0  # [osx]
+  - 1.14.0
```
scopatz: isn't 1.14.4 the most recent build?

croth1: Okay, maybe then I actually did not understand pinning. My thoughts here were:
- I pin 1.14.0 as build time dependency, because it is compatible to 1.14.0 - 1.14.4
- the run time dependency is pinned to 1.14.x in gstreamer's meta.yaml

If I would pin 1.14.4 here, and the runtime dependeny is set to 1.14.x it could be possible to have built with 1.14.4, but having 1.14.1 as run time, which in rare cases can lead to missing symbols.

isuruf: `1.14.4` and `max_pin=x.x` means `>=1.14.4,<1.15.0a0`

croth1: @mingwandroid we actually discussed something very similar recently. Your opinion here may be very valuable.

croth1: @isuruf @scopatz, so it seems that 1.14.4 would work. Shall we go for it then?

isuruf: Sure.

recipe/conda_build_config.yaml
croth1: will do now.

recipe/conda_build_config.yaml
```diff
@@ -332,8 +330,7 @@ gf2x:
...
 gst_plugins_base:
   - 1.12  # [linux]
```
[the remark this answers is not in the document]
croth1: gst_plugin_base seems to be obsolete now. Didn't want to touch it. The only thing that makes sense IMO is removing it completely.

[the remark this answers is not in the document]
isuruf: Isn't it built by `gstreamer-feedstock`?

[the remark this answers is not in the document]
croth1: Ah you're right! Will update this as well then!

recipe/conda_build_config.yaml
```diff
@@ -332,11 +326,9 @@ gf2x:
 gsl:
   - 2.4
 gstreamer:
-  - 1.12  # [linux]
-  - 1.14.0  # [osx]
+  - 1.14.4
 gst_plugins_base:
```
croth1: @isuruf is it ok that it's `gst_plugins_base` here, but `gst-plugins-base` elsewhere? Looks like an inconsistency.

isuruf: That's fine. It's a conda-build thing because these keys are used as jinja variables they should be valid jinja identifiers.

scopatz:
LGTM Thanks for all your hard work @croth1!

croth1:
Thanks to you and @isuruf for helping & giving feedback!

PR #184, opened 2019-02-11
Add librsvg to pinning

henryiii:
This is breaking AfterImage -> ROOT in conda-forge/root-feedstock#15.


scopatz:
Did you want to bump the version number to today too?

henryiii:
Okay, now I know what you mean. Sure! Just do it by hand, then, and PR?

PR #185, opened 2019-02-11
Pinning gdk-pixbuf

henryiii:
This was also seen in conda-forge/root-feedstock#15.


scopatz:
Thanks did you want to bump the version number?

henryiii:
I think so. If you don't mind me asking though, of what?

PR #187, opened 2019-02-11
Typo fixed

henryiii:
This was my typo from a previous PR. Sorry!

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -144,7 +144,7 @@ pin_run_as_build:
   gdal:
     max_pin: x.x
   gdk_pixbuf:
-    max_bin: x.x
+    max_pin: x.x
   geos:
     max_pin: x.x.x
   geotiff:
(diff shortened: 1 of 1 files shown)


henryiii:
Apparently, patch releases of gdk-pixbuf make a difference too. 2.36.9 and 2.36.12 are not compatible. @chrisburr can edit now, I'll lose internet till tomorrow morning soon.

chrisburr:
@scopatz This set of fixes was triggered by https://github.com/conda-forge/root-feedstock/issues/15, in summary installing packages that depend on `librsvg` on macOS can end up with a broken environment. To fix it both `librsvg` and `gdk-pixbuf` must be updated manually.

Am I correct in thinking the pinnings here are only applied to new builds? If so, what is the best way of preventing this from happening to someone else?

1. Fix the channel metadata by making a PR to ` conda-forge-repodata-patches `
2. Rerender and bump the build number for `librsvg` and any packages that are dependent on `librsvg`

scopatz:
Yeah, this only applies to new builds after a rerender

PR #192, opened 2019-02-18
We do not want to support older pythons on the new achitectures

mariusvniekerk:
Checklist

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -481,8 +481,8 @@ poppler:
 proj4:
   - 5.2.0
 python:
-  - 2.7
-  - 3.6
+  - 2.7      # [not (aarch64 or ppc64le)]
+  - 3.6      # [not (aarch64 or ppc64le)]
   - 3.7
 pyqt:
   - 5.6


mariusvniekerk:
cc @jjhelmus @hmaarrfk

hmaarrfk:
i don't know, i guess I kinda don't want to backport the whole python 3.7 build to 3.6, but missing the python 2.7 build might make somethings harder to compile. I guess it is a good time to cleanup that logic.

mariusvniekerk:
My proposal for py27 is just to use yum for it.  We should not be building by27 packages for the new architectures

hmaarrfk:
The beauty of python is that it is an interpretted language. we are enabling people to use new hardware. Python 2.7 compiles well on arch:

https://anaconda.org/archiarm/python/files

I don't think we should be imposing our **bleeding edge** views on others just yet.

mariusvniekerk:
i think we can build py27, we just dont want to build things that are downstream of it.

hmaarrfk:
so you don't want to build matplotlib 2.2.3 for example?

mariusvniekerk:
for py37 sure, why not

hmaarrfk:
truthfully, i don't care about py2.7, py3.6 I think we should build.

At the very least, sometimes py3.7 breaks, and users can quite easily fall back on py.36 if they need.

hmaarrfk:
more importantly, I understand our time is limited, and as you said before, many py2.7 issues will keep comming up.

mariusvniekerk:
@hmaarrfk if you want to port the py36 build we can keep py36, but otherwise probably this.

hmaarrfk:
I don't have time to port 3.6 this next week, so that is why I basically
approve moving forward.

jjhelmus:
Rather than framing this as a "never support older python" I think it might be better to think of this as "python 3.7 first".  If there is interest and people willing to do the work there is no reason why other version of Python cannot be supported on these architectures but getting a Python 3.7 stack out initially seems like a reasonable goal.

jjhelmus:
Python 2.7 and 3.6 are available in `defaults` for the `linux-ppc64le` platform.  Perhaps this filter should be limited to aarch64?

mariusvniekerk:
@jjhelmus for ppc64le i'd personally prefer that its the same as for aarch64 just because it makes things easier as far as the migration goes.  If there is demand we can add the others later.

hmaarrfk:
I just finished cherry-picking all your mods marius. Things seems to be building.

The reason why i don't like skipping 2.7 and 3.6, is that the same issue will pop up with numpy 1.11.

We will need to build basically many numpy versions. It will take a while, but otherwise, we will need to add exceptions to so many pinnings and recipes.

Some software requires numpy 1.13 to build so :/

hmaarrfk:
In 3 hours, it still doesn't work, feel free to merge this. 

I would just prefer if we didn't have needless aarch64 restrictions

hmaarrfk:
Ultimately, I don't want the python 3.6, 3.7, 2.7 recipes to diverge too much. The ability to cherry-pick between recipes is something I don't want to loose. If the arm migration makes things harder for other maintainers, that isn't a good thing.

it is already too complicated to get things into conda-forge due to the multi-os nature of things.

hmaarrfk:
xref: https://github.com/conda-forge/python-feedstock/pull/237
xref: https://github.com/conda-forge/python-feedstock/pull/236

mariusvniekerk:
given that they build now we can cancel this one

PR #193, opened 2019-02-22
readline 8

CJ-Wright:
Checklist

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -493,7 +493,7 @@ pyqt:
 qt:
   - 5.6
 readline:
-  - 7.0
+  - 8.0
 r_base:
   - 3.5.1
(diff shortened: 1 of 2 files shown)


CJ-Wright:
Attn: @conda-forge/core

mariusvniekerk:
So.... this has the potential to break cli application for quite a while.  Are we sure we want to land this without a migrator?

CJ-Wright:
do you want the migrator to land first?

jjhelmus:
How would this break cli applications?  Current packages are build against `readline` 7.0 and should be pinned appropriately.  If they are not they are already broken by the existence of the `readline` 8.0 package.  

As downstream packages are built this will create islands of incompatible packages (those with readline 7 and those with readline 8) but the solver will prevent these from being installed into the same environment.

PR #195, opened 2019-02-22
update build no for openssl

CJ-Wright:
Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -8,7 +8,7 @@ source:
   path: .
 
 build:
-  number: 1 
+  number: 1
   noarch: generic
   script:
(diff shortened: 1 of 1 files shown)


recipe/meta.yaml
```diff
@@ -8,7 +8,7 @@ source:
   path: .
 
 build:
-  number: 0
+  number: 1
```
CJ-Wright:
```suggestion
  number: 1
```

recipe/meta.yaml
```diff
@@ -8,7 +8,7 @@ source:
...
```
ocefpaf: I prefer to always bump the `CalVer` above and just leave the build number `0` forever. It would avoid confusions like this and will always help us spot the version/release date immediately.

PR #211, opened 2019-03-18
pin libspatialindex

ocefpaf:
Goes together with https://github.com/conda-forge/libspatialindex-feedstock/pull/18.

We are not pinning libspatialindex at the moment and everything build with `1.8.5` may be broken once we merge https://github.com/conda-forge/libspatialindex-feedstock/pull/18. I guess we need to do a repo patch in this case. Unless someone has a better idea...


scopatz:
Thanks @ocefpaf!

PR #212, opened 2019-03-19
Update sundials pin

jschueller:
a version 3.2.1 is out, I want to be sure it will be pinned

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -557,7 +557,7 @@ sox:
 sqlite:
   - 3
 sundials:
-  - 3.2 
+  - 3.2.*
 tk:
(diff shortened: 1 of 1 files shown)


recipe/conda_build_config.yaml
```diff
@@ -557,7 +557,7 @@ sox:
 sqlite:
   - 3
 sundials:
-  - 3.2 
+  - 3.2.*
```
ocefpaf: If I understand this correctly the `3.2` pin and `3.2.*` will be the same thing b/c the pin is the minor version (`x.x`), right? If so, can we close this one?

jschueller: yes, I guess so

PR #219, opened 2019-04-11
re-pin geotiff

ocefpaf:
Even though the geotiff 1.4 series was OK with the x.x pin I decided to be more conservative here to avoid headaches. We don't have ABI info on the 1.5.x series.

Pinging @hobu who may have more info on the ABI compatibility. See https://abi-laboratory.pro/?view=timeline&l=libgeotiff for the past info.

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -185,7 +185,7 @@ pin_run_as_build:
   geos:
     max_pin: x.x.x
   geotiff:
-    max_pin: x.x
+    max_pin: x.x.x
   glew:
     max_pin: x.x
   glpk:
@@ -396,7 +396,7 @@ gdal:
 geos:
   - 3.7.1
 geotiff:
-  - 1.4.2
+  - 1.5.1
 giflib:
   - 5.1.7
 glew:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "2019.04.11" %}
+{% set version = "2019.04.12" %}
 
 package:
   name: conda-forge-pinning


ocefpaf:
BTW, I'm planning to merge this as soon as it passes b/c we need this to complete the `proj4` migration.

hobu:
libgeotiff 1.5 will also require proj6 and sqlite3 as part of the https://gdalbarn.com transition. See @erouault blog posts there for exact details. http://osgeo-org.1560.x6.nabble.com/gdal-dev-libgeotiff-1-5-0-release-candidate-td5398253.html also provides some info.

ocefpaf:
> libgeotiff 1.5 will also require proj6 and sqlite3 as part of the [gdalbarn.com](https://gdalbarn.com) transition.

`proj6` is done but we still need to build with `sqlite3`.

Any info on the ABI for the 1.5 series? Is it safe to use major.minor pinning?

hobu:
>  Is it safe to use major.minor pinning?

I'll cautiously say yes? 😄 There was effort done to prevent removal of previously available symbols and no-op them. 

Debian has some tracking on it too.

https://metadata.ftp-master.debian.org/changelogs//main/libg/libgeotiff/libgeotiff_1.5.1-1~exp2_changelog

https://buildd.debian.org/status/package.php?p=libgeotiff&suite=experimental

ocefpaf:
> Debian has some tracking on it too.

Thanks but there is no ABI in there. I guess we need to as [ABI laboratory](https://abi-laboratory.pro/?view=timeline&l=libgeotiff) to track the new source, it seems that they are stuck at `1.4.x`.

> libgeotiff 1.5 will also require proj6 and sqlite3

@hobu I searched the docs and the repo but I could not find how to compile it with `sqlite3`.

hobu:
> @hobu I searched the docs and the repo but I could not find how to compile it with sqlite3.

pkg-config on unix. 

https://github.com/OSGeo/libgeotiff/blob/master/.appveyor.yml describes one way on windows

PR #221, opened 2019-04-24
pin qt

ocefpaf:
~~# DO NOT MERGE YET!!!~~

~~We need to finish https://github.com/conda-forge/qt-feedstock/pull/103 first.~~

That is done! We have `qt 5.9.7` with latest `openssl` for all three main platforms (Windows, Linux, and macOS).


scopatz:
You are a hero @ocefpaf!

PR #236, opened 2019-06-09
Pixman aarch64

hmaarrfk:
xref: https://github.com/conda-forge/cairo-feedstock/pull/48

tjakob
@tjakob
Jun 07 10:41
> Hi all, as part of the migration effort I start to look at cairo (conda-forge/cairo-feedstock#48). It is building fine for POWER, but failing on x86 and aarch64. For x86, change the gcc version to 5.4 makes it work, and the aarch64 build needs pixman 0.38 (as 0.34 is not available)
How would I go about updating the version dependencies?

Checklist

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -531,8 +531,7 @@ pango:
 perl:
   - 5.26
 pixman:
-  - 0.34  # [not (aarch64 or ppc64le)]
-  - 0.38  # [aarch64 or ppc64le]
+  - 0.38
 poppler:
   - 0.67.0
 proj4:
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -531,7 +531,8 @@ pango:
 perl:
   - 5.26
 pixman:
-  - 0.34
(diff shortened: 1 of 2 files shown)


recipe/conda_build_config.yaml
```diff
@@ -531,7 +531,8 @@ pango:
 perl:
   - 5.26
 pixman:
-  - 0.34
+  - 0.34  # [not (aarch64 or ppc64le)]
+  - 0.38  # [aarch64 or ppc64le]
```
ocefpaf: Let's migrate everything to 0.38 then.

hmaarrfk: That would require a migrator right?

ocefpaf: Yep. Those are relatively easy and won't affect your plans to re-render what you need for `aarch64` or `ppc64le` right way.

hmaarrfk:
Follow instructions here

https://conda-forge.org/docs/maintainer/pinning_deps.html#propagate-pin-changes-with-a-migrator

recipe/conda_build_config.yaml
hmaarrfk: ok done.

PR #246, opened 2019-07-10
cfitsio

ocefpaf:
See https://abi-laboratory.pro/?view=timeline&l=cfitsio

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -362,6 +362,8 @@ bzip2:
   - 1
 cairo:
   - 1.16
+cfitsio:
+  - 3.470
 curl:
   - 7.64
 dbus:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "2019.07.04" %}
+{% set version = "2019.07.10" %}
 
 package:
   name: conda-forge-pinning


ocefpaf:
https://github.com/conda-forge/cfitsio-feedstock/pull/18 was merged and now we need this one to ensure the packages that depends on `cfitsio` are properly pinned.

Ping @teake and @pkgw for a review before we merge this one.

ocefpaf:
That was quick :smile:

teake:
Naive question: why does this need to be pinned here as well as in the `run_exports` of `cfitsio`? Wouldn't the latter be sufficient?

Would this also not put a (small) barrier to upgrading `cfitsio`? I.e. I would have to update this feedstock if I want to build against a newer version, right? (I'm asking because I want to build against a future version, which is due this month).

pkgw:
@teake `run_exports` is sort of the low-level mechanism for trying not to break binary compatibility, but the pinning system provides more flexibility in the version constraints, and is also how we coordinate package upgrades across the ecosystem. When a bunch of packages depend on a library, it allows us to standardize their version constraint and automate the process of updating it when we want to upgrade to a new version that may not be ABI compatible. When there are 50 packages that depend on one library that becomes important.

It is true that if you want to move the system to a new version of a pinned package, you have to put in a PR on this repo, but that's pretty easy. (E.g. see the diff associated of this PR.) Also note that we can upgrade the `cfitsio` package itself in the standard way; it's just that if we want other packages to start building against it, we need to update the pin. If this system had been in place before we wouldn't have broken gdal (conda-forge/gdal-feedstock#303).

ocefpaf:
@teake Peter's answer explain it perfectly. Note that, if you are not installing anything that depends on `cfitsio` but only `cfitsio` itself you will get the latest one.

teake:
Thanks for the reply! I'll happily take your word for it that globally pinning packages is important, though I'm afraid I haven't run in to enough fail scenarios to have come the same conclusion by myself. I'll try to wrap my head around it.

It would have been helpful for me if https://conda-forge.org/docs/maintainer/pinning_deps.html#globally-pinned-packages had a section 'The importance of pinning globally' with one or two examples of things that go wrong when not pinning globally. If I come up with a scenario I'll try to add it there.

ocefpaf:
> It would have been helpful for me if conda-forge.org/docs/maintainer/pinning_deps.html#globally-pinned-packages had a section 'The importance of pinning globally' with one or two examples of things that go wrong when not pinning globally. If I come up with a scenario I'll try to add it there.

Indeed. Our docs are still lacking but they are way better than a few months ago :wink: 

Thanks for taking a look at them!

PR #251, opened 2019-07-13
Update Pango to the 1.42 series

pkgw:
The current build of the 1.42.4 is not available on aarch64/ppc64le, but should become available once a PR to add these arches for the new dependency fribidi is merged (conda-forge/fribidi-feedstock#2). This should happen quickly, I hope, but this could conceivably block the addition of the new arches to some packages.

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -524,7 +524,7 @@ openssl:
 openturns:
   - 1.13*
 pango:
-  - 1.40
+  - 1.42
 pari:
   - 2.11
 perl:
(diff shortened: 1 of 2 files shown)


pkgw:
Also: if/when this PR goes through, would the current best practice be to then add an `add_rebuild_successors` line for Pango in `regro/cf-scripts:auto_tick.xsh`?

ocefpaf:
> Also: if/when this PR goes through, would the current best practice be to then add an `add_rebuild_successors` line for Pango in `regro/cf-scripts:auto_tick.xsh`?

Just add a [like these](https://github.com/regro/cf-scripts/blob/e782fbc0cc89f60745b9fc05aa2f5d2f8839e0b0/conda_forge_tick/auto_tick.xsh#L449-L451) and merge it. The conda-forge status page and the migration should start after a few moments. We remove that line when the migration is done.

PR #258, opened 2019-07-23
add pins needed for the linux-armv7l platform

jjhelmus:
Checklist

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -13,6 +13,7 @@ c_compiler:
   # non-standard arches get built with gcc
   - gcc                        # [aarch64]
   - gcc                        # [ppc64le]
+  - gcc                        # [armv7l]
 
   - vs2008                     # [win]
   - vs2015                     # [win]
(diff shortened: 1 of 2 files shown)


scopatz:
This seems reasonable to me.  Is there any reason not to merge it?

xhochy:
Also looks fine to me. I think the only open item will be how the migrator shall look but that is something to be discussed in https://github.com/regro/cf-scripts

CJ-Wright:
@xhochy can you open an issue on cf-scripts? I think we can reuse the existing arch migration things, and maybe bump the object number? This would cause a restart of the entire arch rebuild, which I presume would then catch the new arch @mariusvniekerk what do you think?

jjhelmus:
> I think we can reuse the existing arch migration things, and maybe bump the object number?

Why not create a new migrator for `armv7l`?  The aarch64 + ppc64le migration is nearly complete and it seems wasteful to rebuild everything just to add `armv7l`

scopatz:
I think this is probably also a separate migrator. It is a separate platform. In any event, that discussion  shouldn't block this.

PR #276, opened 2019-09-24
Add proj pin

ocefpaf:
Once we fix all the feedstocks we can remove the `proj4` name from the pinning scheme.

Ping @snowman2


recipe/conda_build_config.yaml
```diff
@@ -310,8 +310,6 @@ pin_run_as_build:
     max_pin: x.x
   poppler:
     max_pin: x.x
-  proj4:
-    max_pin: x.x.x
```
ocefpaf: Both names have `run_exports`.

PR #296, opened 2019-10-13
add ffmpeg migration @hmaarrfk

CJ-Wright:
Closes https://github.com/regro/cf-scripts/pull/648

PR #297, opened 2019-10-13
Improve comments in migrations/example.exyaml

hmaarrfk:
Checklist

### DIFF
--- a/recipe/migrations/example.exyaml
+++ b/recipe/migrations/example.exyaml
@@ -12,5 +12,5 @@ __migrator:
 
 # The name of the feedstock you wish to migrate
 python:
-  - 3.6    # old version
-  - 3.8a1  # new version
+  - 3.6    # list of new versions
+  - 3.8a1  # potentially more than one
--- a/recipe/migrations/example.exyaml
+++ b/recipe/migrations/example.exyaml
@@ -8,7 +8,7 @@ __migrator:
   kind:
     version
   # Only change the migration_number if the bot messes up,
-  # putting this in will cause a complete rerun of the migration
+  # changing this will cause a complete rerun of the migration
   migration_number:
     1
 
--- a/recipe/migrations/example.exyaml
+++ b/recipe/migrations/example.exyaml
@@ -1,3 +1,5 @@
+# To learn more about migrations read CFEP-09
+# https://github.com/conda-forge/conda-forge-enhancement-proposals/blob/master/cfep-09.md
 # The timestamp of when the migration was made
 # Can be obtained by copying the output of
 # python -c "import time; print(f'{time.time():.0f}')"
--- a/recipe/migrations/example.exyaml
+++ b/recipe/migrations/example.exyaml
@@ -5,7 +5,7 @@ migrator_ts: 0
 __migrator:
   kind:
     version
-  # Only use the migration_number if the bot messes up,
+  # Only change the migration_number if the bot messes up,
   # putting this in will cause a complete rerun of the migration
   migration_number:
     1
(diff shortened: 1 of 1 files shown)


recipe/migrations/example.exyaml
```diff
@@ -1,9 +1,16 @@
...
+  # Only use the migration_number if the bot messes up,
+  # putting this in will cause a complete rerun of the migration
+  migration_number:
+    1
+
+# The name of the feedstock you wish to migrate
 python:
-  - 3.6
-  - 3.8a1
+  - 3.6    # old version
```
CJ-Wright: This is not necessarily true, in this case we're building 3.6 and 3.8a1. The value here is whatever you want the resulting config to be.

recipe/migrations/example.exyaml
```diff
@@ -1,9 +1,16 @@
...
+# The timestamp of when the migration was made
+# Can be obtained by copying the output of
+# python -c "import time; print(f'{time.time():.0f}')"
+migrator_ts: 0
+__migrator:
+  kind:
     version
-  migration_number:  # Only use this if the bot messes up, putting this in will cause a complete rerun of the migration
-    1 
+  # Only use the migration_number if the bot messes up,
```
CJ-Wright:
```suggestion
  # Only change the migration_number if the bot messes up,
```

CJ-Wright:
you might add an xref to https://github.com/conda-forge/conda-forge-enhancement-proposals/blob/master/cfep-09.md

recipe/migrations/example.exyaml
```diff
@@ -1,9 +1,18 @@
...
+  # Only change the migration_number if the bot messes up,
+  # putting this in will cause a complete rerun of the migration
+  migration_number:
+    1
+
+# The name of the feedstock you wish to migrate
 python:
-  - 3.6
-  - 3.8a1
+  - 3.6    # list of new versions
```
CJ-Wright:
```suggestion
  - 3.6    # list of versions to build against
```

hmaarrfk:
Where does the new pinning get set?

CJ-Wright:
What do you mean exactly?
Conda smithy has the ability to merge all the pinning migration yaml files and the global config before it creates the variants so that the effective config is the composite.

For the most part the values in the pinning yaml override the values in the global config where they are in conflict. In the example we update the python versions we build against to 3.6 and 3.8a1 since we specified the python key and gave those two versions. For a simple version bump migration one would provide a yaml with the new version, like done in #296

hmaarrfk:
Right, it is just unclear to me what a list of versions means for a pinning. But I kinda get it

CJ-Wright:
A list of versions mean that we build against all the listed versions as a build matrix for instance [python](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/master/recipe/conda_build_config.yaml#L545)

CJ-Wright:
I updated the example to use boost rather than python, since single version migrations are much more common than build matrix migrations. The full spec for running complex migrations is in the CFEP9 proposal itself.

recipe/migrations/example.exyaml
```diff
@@ -1,9 +1,17 @@
...
+__migrator:
+  kind:
     version
-  migration_number:  # Only use this if the bot messes up, putting this in will cause a complete rerun of the migration
-    1 
-python:
-  - 3.6
-  - 3.8a1
+  # Only change the migration_number if the bot messes up,
+  # putting this in will cause a complete rerun of the migration
```
CJ-Wright:
```suggestion
  # changing this will cause a complete rerun of the migration
```

CJ-Wright:
Thank you very much!

hmaarrfk:
No problem. Thanks for your help!

PR #301, opened 2019-10-17
migrator for mpich 3.3

minrk:
this version is already pinned in conda_build_config.yaml, but several packages were not rebuilt due to multiple outputs (e.g. ptscotch). I'm not sure if the new migrators work when the new version is already in the pins, even though there are still packages that need rebuilding, but this should tell us.

### DIFF
--- a/recipe/migrations/mpich-3.3.yaml
+++ b/recipe/migrations/mpich-3.3.yaml
@@ -1,0 +1,17 @@
+# To learn more about migrations read CFEP-09
+# https://github.com/conda-forge/conda-forge-enhancement-proposals/blob/master/cfep-09.md
+# The timestamp of when the migration was made
+# Can be obtained by copying the output of
+# python -c "import time; print(f'{time.time():.0f}')"
+migrator_ts: 1571312255
+__migrator:
+  kind:
+    version
+  # Only change the migration_number if the bot messes up,
+  # changing this will cause a complete rerun of the migration
+  migration_number:
+    1
+
+# The name of the feedstock you wish to migrate
+mpich:
+  - 3.3    # new version to build against


recipe/meta.yaml
```diff
@@ -1,4 +1,4 @@
-{% set version = "2019.10.11" %}
+{% set version = "2019.10.17" %}
```
CJ-Wright: FYI yo don't need to update the version number here, since we aren't updating the global pinnings, just starting the migration.

CJ-Wright:
Otherwise LGTM

recipe/meta.yaml
minrk: Updated to remove the version bump, thanks! I'm not 100% clear what information the migrators operate on. This should still work, even though the global pinnings already have this version?

CJ-Wright: Yes, I think so. At the very least they should get a rebuild PR (even if the actual migration yaml is moot).

CJ-Wright:
I'm going to merge this, but it may take a while for the bot to kick in. The new version of networkx has been hitting the bot hard.

minrk:
Looking at the migrator status, the new migrator missed scotch and mumps and . Scotch was manually updated, so that may be why, but mumps hasn't (A PR was issued manually in an attempt to rebuild, but it was merged too early, and it was built with 3.2).

I think the fact that mpi providers are selected with an `{{ mpi }}` variable is somehow causing packages to be omitted from the migrator. Is that possible? What's a good way to test the migrator finding the right feedstocks in the graph?

minrk:
I also noticed that it shows petsc as 'done' in the graph, which is not true. petsc is built with mpich 3.2, and didn't receive a PR.

PR #302, opened 2019-10-17
package license_file in this feedstock

minrk:
fixes linter errors about missing license_file


minrk:
hm, builds fine locally, but apparently not with docker. Tried to copy conda-smithy https://github.com/conda-forge/conda-smithy/blob/master/conda_smithy.recipe/meta.yaml

PR #306, opened 2019-10-21
Disable python 2.7 for ppc64le and aarch64

hmaarrfk:
We didn't build scipy for python 2.7
It seems that we are running the clock on 2019 anyway. 
Marius would have called this back in 2018 when we started this project ;)

cc: @mariusvniekerk 
@conda-forge/arm-arch 

Checklist

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -547,7 +547,7 @@ proj4:
 proj:
   - 6.2.0
 python:
-  - 2.7
+  - 2.7  # [not (aarch64 or ppc64le)]
   - 3.6
   - 3.7
 qt:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "2019.10.11" %}
+{% set version = "2019.10.21" %}
 
 package:
   name: conda-forge-pinning


jakirkham:
Yeah we are finding ourselves more often than not needing to drop Python 2.7 for these architectures manually anyways.

Thoughts @conda-forge/core?

jjhelmus:
Is it worth dropping Python 2.7 from these platforms when in two months Python 2 is going to be dropped for all platforms?  I do not have a strong opinion either way but think it is worth consideration.

jakirkham:
> Is it worth dropping Python 2.7 from these platforms when in two months Python 2 is going to be dropped for all platforms?

I think the challenge is people are having to go through and manually fix arch migrators to drop Python 2 and then wait for feedback from maintainers about whether this is ok. This would generally let people know we are ok with dropping Python 2.7 on the migrators.

> I think it is probably OK to stop building Python 2.7 for all platforms by default now. Folks can manually re-enable it if they need to.

Personally I wouldn't go quite that far. Though I definitely think we should drop it the day the Python core group does.

CJ-Wright:
RE: the migrators
We could update the migration code to drop py2.7 going forward. Then maintainers can ask for a bot-rerun and they should get the py2.7 drop in the new PR.

hmaarrfk:
> We could update the migration code to drop py2.7 going forward. Then maintainers can ask for a bot-rerun and they should get the py2.7 drop in the new PR.

This is what this PR proposes right?

CJ-Wright:
Yes, sorry I didn't read the source 🐑

hmaarrfk:
To address jjhelmus question, I'm not too interested in making policy decisions at this stage, but. Aarch64 only built newer packages that mostly already dropped python 2.7. As such. There are real holes in the build matrix. This PR just avoids tedious fixes and back and forth.

jakirkham:
> There are real holes in the build matrix. This PR just avoids tedious fixes and back and forth.

That was my understanding as well.

mariusvniekerk:
yeah building py27 things for these arches is way more trouble than it is worth

PR #310, opened 2019-10-24
Pin `cudnn` globally

jakirkham:
The `cudnn` package shipped by `defaults` is batteries included. Though it currently does not have `run_exports`. To establish some kind of pinning for the `defaults` package of `cudnn` when used by other packages, this sets the `cudnn` version to 7. Also it sets the `min_pin` to respect forward and backward compatibility of the library at the patch version. Finally as the library maintains backwards compatibility for each major release, this sets `max_pin` to respect that.

Checklist


jakirkham:
Thoughts @conda-forge/core?

CJ-Wright:
We only ever had the one version, right?

jakirkham:
Yeah I think developers and users are going to always want to be on the latest version with this one.

As they established API/ABI compatibility guidelines in version 7 (included here), I think they are unlikely to break those anytime soon.

CJ-Wright:
I mostly wanted to make sure that we didn't need to bring everyone *up* to this version.

recipe/conda_build_config.yaml
```diff
@@ -356,6 +359,8 @@ cairo:
   - 1.16
 cfitsio:
   - 3.470
+cudnn:
+  - 7
```
CJ-Wright: If the min pin is `x.x` do we need the pinned version to `x.x`?

jakirkham: Well this is more-or-less the same discussion we are having in [this thread]( https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/309#pullrequestreview-305533073 ). Namely do we want a policy of tightly pinning the minimum or not?

With `cudnn` I would suspect users, developers, and maintainers are eager to get the latest version in their builds for the latest optimizations. So kept this loose.

With `nccl` I'm guessing users are more concerned that things stay operational.

I could be wrong about both of these guesses. As a community we may decide that other things are more important than our thoughts on user preferences.

Kind of a side note, but the default `min_pin` is `x.x.x.x.x.x.x`.

PR #318, opened 2019-11-01
use bump no so python doesn't rebuild everything

CJ-Wright:


CJ-Wright:
@conda-forge/core this has hooks for not bumping the build number for new python versions, merging it without upstream should be ok. It's just metadata that the bot would otherwise ignore.

PR #320, opened 2019-11-03
Protobuf

ocefpaf:
See https://github.com/conda-forge/protobuf-feedstock/issues/75

### DIFF
--- a/recipe/migrations/protobuf310.yaml
+++ b/recipe/migrations/protobuf310.yaml
@@ -11,7 +11,8 @@ __migrator:
   # changing this will cause a complete rerun of the migration
   migration_number:
     1
-
+  build_number:
+    1
 # The name of the feedstock you wish to migrate
 libprotobuf::
   - 3.10
--- a/recipe/migrations/protobuf310.yaml
+++ b/recipe/migrations/protobuf310.yaml
@@ -1,0 +1,17 @@
+# To learn more about migrations read CFEP-09
+# https://github.com/conda-forge/conda-forge-enhancement-proposals/blob/master/cfep-09.md
+# The timestamp of when the migration was made
+# Can be obtained by copying the output of
+# python -c "import time; print(f'{time.time():.0f}')"
+migrator_ts: 1572797096
+__migrator:
+  kind:
+    version
+  # Only change the migration_number if the bot messes up,
+  # changing this will cause a complete rerun of the migration
+  migration_number:
+    1
+
+# The name of the feedstock you wish to migrate
+libprotobuf::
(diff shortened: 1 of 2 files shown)


CJ-Wright:
Can we hold off on this for a little bit? I need to fix the build bump stuff

ocefpaf:
> Can we hold off on this for a little bit? I need to fix the build bump stuff

Sure. There is no rush. I can also build the older version with py38.

ocefpaf:
> Can we hold off on this for a little bit? I need to fix the build bump stuff

@CJ-Wright I think this should be OK now, right?

CJ-Wright:
The issue is this line: https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/migrators.xsh#L1325 we set the build number bump to zero for the python migration. We need to set it back to 1 for this migration, or get this set from the migration yaml itself. Part of https://github.com/regro/cf-scripts/pull/689/files handled this.

recipe/migrations/protobuf310.yaml
```diff
@@ -0,0 +1,18 @@
...
+migrator_ts: 1572797096
+__migrator:
+  kind:
+    version
+  # Only change the migration_number if the bot messes up,
+  # changing this will cause a complete rerun of the migration
+  migration_number:
+    1
+  build_number:
+    1
```
ocefpaf: @CJ-Wright is this correct?

CJ-Wright: yes

ocefpaf: So, merge :grimacing:

CJ-Wright: No, not yet need to fix bot first

ocefpaf:
> This is pending a fix to the bot

No problem. I thought the issue was with this PR. I'll try to create an older build of `protobuf` instead to unblock some of the py38 migration PRs.

CJ-Wright:
Yeah the issue here is that we can't configure the build number stuff. I'm working on fixing it.

ocefpaf:
> Yeah the issue here is that we can't configure the build number stuff. I'm working on fixing it.

No rush. We actually may need the protobuf stuff for 3.8 anyways.

CJ-Wright:
I want to get this in hand. It is very frustrating that we can't run these migrations side by side.

ocefpaf:
> I want to get this in hand. It is very frustrating that we can't run these migrations side by side.

I know :smile:, just no pressure from my side. I own you way too many beers already.

PR #322, opened 2019-11-05
add build number to example

CJ-Wright:


recipe/migrations/example.exyaml
```diff
@@ -11,6 +11,11 @@ __migrator:
   # changing this will cause a complete rerun of the migration
   migration_number:
     1
+  # This determines the increment to the build number when the 
+  # migration runs. 
+  # Change this to zero if the new pin increases the number of builds
+  build_number:
+    1
```
jakirkham: What is the default if this is not set?

ocefpaf: Yep. The default should be to bump and leaving it untouched the special case.

CJ-Wright: Yes. I just wanted to include it in case people wanted to know about the special case

PR #328, opened 2019-11-14
migrator for petsc/slepc 3.12

minrk:


recipe/migrations/petsc-3.12.yaml
```diff
@@ -0,0 +1,23 @@
+# To learn more about migrations read CFEP-09
+# https://github.com/conda-forge/conda-forge-enhancement-proposals/blob/master/cfep-09.md
+# The timestamp of when the migration was made
+# Can be obtained by copying the output of
+# python -c "import time; print(f'{time.time():.0f}')"
+migrator_ts: 0\1573720464
```
CJ-Wright: I don't think this is correct.

minrk: Fixed, thanks!

CJ-Wright:

PR #347, opened 2019-11-27
Revert "Change python order to give higher versions priority"

mbargull:
This reverts commit 4ab7be19bc03255fcf90d1181caf05f55603dcaf.

Migrations don't place nice with this, unfortunately


mbargull:
How do we handle same day version changes? `2019.11.27.1` would make most sense, right?

mbargull:
Or just move `2019.11.27` to `broken`.

jakirkham:
We’ve been doing build number bumps for additional same day releases.

PR #357, opened 2019-12-17
finish out libnetcdf

ocefpaf:
This is safe to be merged b/c the remaining packages on the migration are probably too broken to build.


ocefpaf:
Thanks!

CJ-Wright:
Sure, although we should note those packages which we consider "too broken to build" somehow

ocefpaf:
> Sure, although we should note those packages which we consider "too broken to build" somehow

On it in their feedstock. We'll probably archive them but it is pending the team's decision.

recipe/conda_build_config.yaml
```diff
@@ -224,8 +224,6 @@ pin_run_as_build:
     max_pin: x.x
   libmatio:
     max_pin: x.x
-  libnetcdf:
-    max_pin: x.x.x
```
xylar: Was there a discussion that it's no longer necessary to pin libnetcdf build as run?  I would have thought this needed to be at least `x.x`.  That was my understanding of this comment: https://github.com/conda-forge/libnetcdf-feedstock/pull/90#issuecomment-558239517

xylar: Maybe it's fine.  https://abi-laboratory.pro/index.php?view=timeline&l=netcdf shows 100% backwards compatibility for all the releases we're likely to have on conda-forge.

ocefpaf: @xylar with `run_exports` on the feedstock we don't need this here but the same pin still applies.

PR #387, opened 2020-01-31
Migrate `astroid`

jakirkham:
Guess we were missing this one even though it is a `pylint` dependency 🤷‍♂

cc @nerohmot

Checklist

### DIFF
--- a/recipe/migrations/arch_rebuild.txt
+++ b/recipe/migrations/arch_rebuild.txt
@@ -76,3 +76,4 @@ spyder-kernels
 pyvisa-py
 opcua
 ipykernel
+astroid


CJ-Wright:
Yeah we need to fix the graph a little bit.

jakirkham:
I'm guessing this is because we are not looking at dependencies of `noarch` packages. Is that right or is this a different issue?

CJ-Wright:
Not certain, we pluck `noarch` pacakges, which removes them from the graph but should preserve their edges. We may not be doing that properly. There also could be other errors in the filtering logic.

Relevant bits: 
https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/auto_tick.py#L295
https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/migrators/arch.py#L39

Update:
I think I found some of the issues
1. I think we can remove https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/auto_tick.py#L306 since this might be handled by the init
2. https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/migrators/arch.py#L65 most likely should be pluck and should include the noarch condition.

@mariusvniekerk what do you think?

CJ-Wright:
This is in there it's awaiting wrapt

jakirkham:
Good catch! Thanks CJ 😄

I must have looked into that before as I asked the bot to re-run the migration on PR ( https://github.com/conda-forge/wrapt-feedstock/pull/14 ), but it hasn't happened yet. Guessing this is due to the same known issue; more motivation for this work ( https://github.com/conda-forge/conda-forge-webservices/issues/289 ) it seems 🙂

scopatz:
This has conflicts now

CJ-Wright:
I think this can be closed?

jakirkham:
IIUC @CJ-Wright we can just close this. Is that right?

jakirkham:
Actually completely missed PR ( https://github.com/conda-forge/astroid-feedstock/pull/36 ) is already here to migrate `astroid`. Closing this out. Thanks all! 😄

PR #392, opened 2020-02-05
Add migration for gsl=2.6

xhochy:
ABI breaking, so we need to migrate: https://abi-laboratory.pro/?view=timeline&l=gsl


xhochy:
@conda-forge/core Please review/merge.

PR #405, opened 2020-02-24
migrate to proj 6.3.1

ocefpaf:
We still have a tight `x.x.x` pin due to https://abi-laboratory.pro/?view=timeline&l=proj. Hopefully we'll be able to relax that pin with `proj 7.0`.


scopatz:
Thanks!

PR #410, opened 2020-03-02
Add migration for root=6.20.0

duncanmmacleod:
cc @chrisburr.

Checklist

### DIFF
--- a/recipe/migrations/root620000.yaml
+++ b/recipe/migrations/root620000.yaml
@@ -1,0 +1,15 @@
+migrator_ts: 1583150223
+__migrator:
+  kind:
+    version
+  migration_number:
+    1
+  build_number:
+    1
+
+root:
+  - 6.20.0
+root_base:
+  - 6.20.0
+root_binaries:
+  - 6.20.0


chrisburr:
Most of the repos I expected to be picked up by this haven't been. Am I just being impatient or is something wrong? It seems to be related to how the multiple outputs are used but I don't see what could be confusing the bot.

In summary `root` is a meta package which depends on `root_base`. Only packages which explicitly depend on `root` have been migrated.

Examples:

 * https://github.com/conda-forge/root_numpy-feedstock/blob/master/recipe/meta.yaml
 * https://github.com/conda-forge/lhcb-simpletools-feedstock/blob/master/recipe/meta.yaml

xhochy:
> Most of the repos I expected to be picked up by this haven't been. Am I just being impatient or is something wrong? It seems to be related to how the multiple outputs are used but I don't see what could be confusing the bot.
> 
> In summary `root` is a meta package which depends on `root_base`. Only packages which explicitly depend on `root` have been migrated.

That's expected. If you also want to have migrations for `root_base`, you need to add it to the migration. See the `boost` and `boost-cpp` split.

chrisburr:
@xhochy I'm still not sure what's wrong as both `root` and `root_base` are already included in this PR. It seems to be the same as the boost migration.

PR #412, opened 2020-03-03
Fix root-binaries package name in migration and pinning list

duncanmmacleod:
This PR corrects the package name for `root-binaries`, which was previously incorrectly listed as `root_binaries` in both the root-62000 migration, and the `conda_build_config.yaml` pinning list.

cc @chrisburr

Checklist

### DIFF
--- a/recipe/migrations/root620000.yaml
+++ b/recipe/migrations/root620000.yaml
@@ -3,7 +3,7 @@ __migrator:
   kind:
     version
   migration_number:
-    2
+    1
   build_number:
     1
 
--- a/recipe/migrations/root620000.yaml
+++ b/recipe/migrations/root620000.yaml
@@ -3,7 +3,7 @@ __migrator:
   kind:
     version
   migration_number:
-    1
+    2
   build_number:
     1
 
@@ -11,5 +11,5 @@ root:
   - 6.20.0
 root_base:
   - 6.20.0
-root_binaries:
+root-binaries:
   - 6.20.0
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -571,7 +571,7 @@ root:
   - 6.18.04
 root_base:
   - 6.18.04
-root_binaries:
+root-binaries:
   - 6.18.04
 ruby:
   - 2.5
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "2020.02.20" %}
+{% set version = "2020.03.03" %}
 
 package:
   name: conda-forge-pinning


recipe/migrations/root620000.yaml
```diff
@@ -3,13 +3,13 @@ __migrator:
   kind:
     version
   migration_number:
-    1
+    2
```
beckermr: This will force the migration to start over and reissue PRs even for things it has already done. Is that what you want?

beckermr: If the migration never did anything to begin with, you might keep this the same.

chrisburr: No, the two which have already been issued are okay.
```suggestion
    1
```

duncanmmacleod: I have no idea, I just followed the example from #403. I think @chrisburr has followed the progress of this migration and can maybe comment?

beckermr: Right. Boost was done wrong there IIRC so yeah let's keep it the same.

duncanmmacleod: This migration has only discovered one (1) package that needed migrating (which isn't correct anyway), so opening new PRs isn't going to be a big deal.

chrisburr:
Would this have been the cause of PRs not being issued for packages that depend on `root_base` and `root-binaries`? (see https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/410#issuecomment-593461541)

I expected them to get picked up anyway due to `root_base` being listed, but maybe the scripts got confused?

beckermr:
The "-" is causing issues. @duncanmmacleod @chrisburr

recipe/migrations/root620000.yaml
chrisburr: It actually discovered two but one got lost after it was merged: https://github.com/conda-forge/rapidsim-feedstock/pull/2/files

duncanmmacleod:
FFS, someone's already been over this ground: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/358/commits/9a7d26dec2239d9ee92f6e0a90a2bc752a3a4afb

We can probably just close this and ignore - it still doesn't explain the poor migration discovery.

beckermr:
🤦‍♂ 

@duncanmmacleod let's open an issue on regro/cf-scripts. This looks like an error in the bot graph of some sort.

chrisburr:
The only reason the dash is there is to avoid a bug in conda-build where it builds `root-binaries` and `root_base`...

It's okay as we don't actually need the `root-binaries` pin as everything that includes it will also have `root_base` pinned to the same version.

PR #415, opened 2020-03-03
Update root620000 migration to only update root_base

duncanmmacleod:
This PR updates the root620000 migration (#410) to only update the `root_base` package (see #414)

cc: @chrisburr @beckermr 

Checklist

### DIFF
--- a/recipe/migrations/root620000.yaml
+++ b/recipe/migrations/root620000.yaml
@@ -7,9 +7,5 @@ __migrator:
   build_number:
     1
 
-root:
-  - 6.20.0
 root_base:
-  - 6.20.0
-root_binaries:
   - 6.20.0


recipe/migrations/root620000.yaml
```diff
@@ -1,15 +1,11 @@
-migrator_ts: 1583150223
+migrator_ts: 1583251649
 __migrator:
   kind:
     version
   migration_number:
-    1
+    2
```
beckermr: We can keep the timestamp and migration number the same I think. 

@chrisburr for double checking this

chrisburr: Yes we can

chrisburr:
@beckermr It's still not working, any ideas?

beckermr:
Nope. We’ll need to dig in the graph.

beckermr:
This might be root_base versus root-base

chrisburr:
Does the pin have to already be rendered into the feedstock's configuration?

I've just realised the pin was added after root_numpy was last rerendered: https://github.com/conda-forge/root_numpy-feedstock/blob/master/.ci_support/linux_python3.8.yaml

beckermr:
The package needs to be in the metadata of the graph. So if the bot is looking for `root_base` but it is denoted as `root-base`, then the bot won't find any feedstocks to migrate. This metadata is parsed out of the meta.yaml files, so my guess is that this is the issue. 

Can you bump us in the bot gitter room? Others can confirm there.

chrisburr:
Where is `root-base` used? It doesn't exist on Anaconda.org and in recipes I can only find instances of `root_base`?

beckermr:
Ohhhhhh. Well then this is a bug.

chrisburr:
I've tried asking the bot to rerender in an issue but it hasn't responded: https://github.com/conda-forge/root_numpy-feedstock/issues/8

Doing it by hand it works so it's not an conda-smithy issue. As I don't think I have access, can you check the logs in case that explains it?

beckermr:
issue rerenders are broken sorry.

beckermr:
https://github.com/conda-forge/conda-forge-webservices/issues/326

PR #423, opened 2020-03-05
End ROOT 6.20.00 migration

chrisburr:
cc @duncanmmacleod


beckermr:
update the date to today and we can merge

PR #430, opened 2020-03-13
BUG exclude cpython only package

beckermr:
We currently don't have good metadata on which packages are cpython only. For now, it may make sense to exclude them by hand.

Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "2020.03.12" %}
+{% set version = "2020.03.13" %}
 
 package:
   name: conda-forge-pinning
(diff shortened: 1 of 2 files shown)


beckermr:

beckermr:
cc @isuruf @chrisburr @synapticarbors

xhochy:
This doesn't seem like a scalable solution for me as there will be tons of incompatible packages, e.g https://github.com/conda-forge/protobuf-feedstock/pull/84 is also failing.

beckermr:
Agreed @xhochy but I am unsure of what else to do. Do you think you could write a migrator to find all packages that are incompatible and cpython? Not all cpython packages will fail with pypy as it has partial support.

xhochy:
I don't think we can detect that. Maybe we should rather check whether packages for PyPy were built before continuing with the migration. But sadly I don't have enough migrator code knowledge to assess the feasibility of implementing this.

beckermr:
Well if a package doesn’t build and we don’t merge the pr then everything will be fine. 

I think we need to generate the metadata which means doing as I propose here. As packages fail, we did their metadata and exclude them. Then we can push a change to the bot using the fixed metadata.

beckermr:
closing per discussion on gitter - if a bad PR is issued we should simply keep it open

PR #494, opened 2020-03-19
end geos migration

ocefpaf:
All PRs were issued.


beckermr:
@ocefpaf we forgot to delete the migration file I will submit another PR.

PR #541, opened 2020-04-10
datetime version, don't merge until tomorrow since we already released tomorrow

CJ-Wright:


CJ-Wright:
I think this is ready now, we successfully uploaded the ci-setup package

beckermr:
We need to wait for the CDN and then we are good. So merge in an hour or so

PR #572, opened 2020-04-23
fix migration

CJ-Wright:
CC @becker


CJ-Wright:
@isuruf this is the fix for the bad bot close out PR. The plan is to fix the migrator, and then ask the bot to rerun on the pinnings PR.

PR #638, opened 2020-06-03
BUG close extra poppler migrations

beckermr:
We have had more than one of these going at once which is a bad idea.


beckermr:
@conda-forge/bot ok to merge?

PR #649, opened 2020-06-07
Build attr & libcap for arm/ppc

saraedum:
see https://github.com/conda-forge/attr-feedstock/pull/1 & https://github.com/conda-forge/libcap-feedstock/pull/10


saraedum:
This is the first time I propose a change to this file. I hope it's enough to just add some lines here.

PR #653, opened 2020-06-11
Pin alsa-lib to 1.1.5, the widely-used version.

ryanvolz:
`alsa-lib` already has a run_exports with max_pin=x.x.x, and up until two days ago there was only the one version released on conda-forge. This will set the pin for that version with no required rebuilds. Then packages can update to the new version of `alsa-lib` with a migrator to follow.

Checklist

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -334,6 +334,8 @@ blas_impl:
 
 abseil_cpp:
   - '20200225.2'
+alsa_lib:
+  - 1.1.5
 arb:
   - 2.17
 arpack:


recipe/conda_build_config.yaml
```diff
@@ -334,6 +334,8 @@ blas_impl:
 
 abseil_cpp:
   - '20200225.2'
+alsa-lib:
```
beckermr, line 337: You need an underscore here, not a dash. Idk why but this is magic we must use.

ryanvolz: Done!

PR #658, opened 2020-06-12
Pin uhd to 3.15.0.

ryanvolz:
There are now a few packages that build against `uhd`'s C/C++ library:
- gnuradio-osmosdr
- gnuradio-uhd (gnuradio parent package)
- soapysdr-module-uhd

I'm not aware of any more packages that might depend on `uhd` in the future, but it would still be nice to have a pin so that keeping these in sync is easier. `uhd` already has a `run_exports` with
`max_pin=x.x.x`.

Checklist

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -602,6 +602,8 @@ tk:
   - 8.6                # [not ppc64le]
 tiledb:
   - 1.7
+uhd:
+  - 3.15.0
 vc:                    # [win]
   - 14                 # [win]
 vlfeat:


recipe/conda_build_config.yaml
```diff
@@ -602,6 +602,8 @@ tk:
   - 8.6                # [not ppc64le]
 tiledb:
   - 1.7
+uhd:
+  - 3.15.0
```
scopatz, line 606: can it just be `3.15`?

ryanvolz: The `run_exports` pin is `x.x.x`, which matches the SOVERSION on the uhd library. So I think that means the third part is necessary, unless `3.15` would imply the `.0` in this case?

scopatz: yep OK

PR #689, opened 2020-07-15
added windows specific graphviz pin

### LINKED ISSUE #688: graphviz pin 2.40
graphviz is pinned to `2.40`. However, the latest windows build is `2.38.0`. Windows with graphviz as a dependency will not build on windows right now

e.g. https://github.com/conda-forge/lemon-feedstock/pull/5, with [corresponding build issue](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=184168&view=logs&j=f97ca392-d626-52d0-4b2e-4b27aa319a77&t=7b750183-140d-50ae-3ee1-cbef6218bdf6&l=62)

```
conda.exceptions.ResolvePackageNotFound: 
  - graphviz=2.40
```

https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/9898f7005d1aabb811e4aba38ece5b3003f4f160/recipe/conda_build_config.yaml#L413-L414

as far as solutions go, it's probably
 * to add a `[unix]` selector after the `2.40` one and adding `2.38` with `[win]`, or
 * to obtain a current build on windows as well...

k-dominik:
Checklist

fixes #688 

added windows specific graphviz pin to the latest available version `2.38` (the rest of the platforms already moved on to `2.40`).
Without the windows specific pin, windows recipes relying on graphviz will not build  because this dependency constraint cannot be satisfied.

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -411,7 +411,8 @@ google_cloud_cpp_common:
 googleapis_cpp:
   - '0.10'
 graphviz:
-  - 2.40
+  - 2.38  # [win]
+  - 2.40  # [not win]
 grpc_cpp:
   - '1.30'
 harfbuzz:


beckermr:
Can you explain this?

k-dominik:
> added windows specific graphviz pin to the latest available version `2.38` (the rest of the platforms already moved on to `2.40`).
> Without the windows specific pin, windows recipes relying on graphviz will not build because this dependency constraint cannot be satisfied.

Hi @beckermr , you're quick... I added the above sentence to explain the PR.

beckermr:
Ahhh I see. @conda-forge/graphviz why is 2.40 not available on windows?

mbargull:
refs:
@mbargull in https://github.com/conda-forge/lemon-feedstock/pull/14#discussion_r454692485 :
> I'm no expert on Graphviz at all. But this should be fine given:
> - the [`max_pin: x` in `conda-forge-pinning`](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/cc418d28617f0581baf74a7f0feb73c7f2c57844/recipe/conda_build_config.yaml#L188-L189)
> - and the track record from https://abi-laboratory.pro/?view=timeline&l=graphviz .

@jakirkham in https://github.com/conda-forge/lemon-feedstock/pull/14#discussion_r454698890 :
> Yeah this is the right thing to do I think. We are just doing binary repackaging on Windows anyways (due to the complexity of the Graphviz build system). Also `2.38.0` was the last binary they uploaded for Windows. Maybe this will be easier to build from source with CMake support they added ( https://github.com/conda-forge/graphviz-feedstock/issues/11 ). All that to say this seems like a good solution for now. 🙂

----
It seems building Graphviz on Windows could be a cumbersome undertaking. I mean it **should be done**, of course.
But blocking downstream doesn't seem reasonable due to `max_pin: x` and API stability (acc. to abi-laboraroy.pro).

beckermr:
@CJ-Wright do we need to do a migration here for the downgraded pin?

mbargull:
I don't think we need a migration. I just manually checked the few PRs from https://github.com/search?l=&q=Rebuild+for+graphviz-2.40++user%3Aconda-forge&type=Issues and most are either `noarch: python` or have `skip: True  # [win]` anyway. Apart from `lemon` (which this PR is meant to help with), the exceptions are only `pygraphviz` which pins to `2.38.*` in its `meta.yaml` anyway and `ibis-framework` which has a `run`-only dep on `graphviz`.

@beckermr, if you concur, feel free to merge.

mbargull:
(And I only checked those PRs to see if someone added workarounds because of failed Windows builds -- we don't really downgrade because we never had `2.40` on Windows.)

CJ-Wright:
I think we can get away without the downgrade migration, since windows never built with a more recent version, but if we start seeing issues we can come back to this.

mbargull:
@k-dominik, thanks for the PR and everyone else for the review!

PR #760, opened 2020-09-08
migrate to proj 7.1.1

ocefpaf:
Ping @xylar who I believe is hitting some road blocks b/c of the current pin.


xylar:
I think this isn't a bad idea but I'm not sure if the `rasterio` issue we're seeing will be fixed by this migration.  Could be but hard to be sure.

ocefpaf:
@beckermr or @mariusvniekerk, this one is mostly harmless and good to go.

beckermr:
here goes nothing!

PR #861, opened 2020-10-13
end arrow-cpp

ocefpaf:
Looks like all PRs were issued.


jakirkham:
Looks like PR ( https://github.com/conda-forge/snowflake-connector-python-feedstock/pull/54 ) is still open. Unsure whether we should wait for that or not.

beckermr:
Once the PRs have been issued we can generally close the migration. The only thing we lose is the ability for the bot to remake PRs if needed.

PR #873, opened 2020-10-16
Migrate glib to 2.64.6

xhochy:
This introduces the `libglib` package and will remove the `python` run dependency from a lot of packages, most notably `r-base`.

Checklist

### DIFF
--- a/recipe/migrations/glib2646.yaml
+++ b/recipe/migrations/glib2646.yaml
@@ -9,3 +9,5 @@ __migrator:
 
 glib:
   - 2.64.6
+libglib:
+  - 2.64.6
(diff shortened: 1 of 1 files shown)


recipe/migrations/glib2646.yaml
```diff
@@ -0,0 +1,11 @@
+migrator_ts: 1
+__migrator:
+  kind:
+    version
+  migration_number:
+    1
+  build_number:
+    1
+
+glib:
```
beckermr, line 10: What is the right pin here? Do we need a pin both libglib and glib?

xhochy: I  would expect that for building you always need glob installed. We can pin both to be sure but we shouldn't need to.

PR #895, opened 2020-10-26
Update openturns pin

jschueller:
supercedes #887, all dependent feedstocks are already migrated


xhochy:
@jschueller I'm a bit confused as #887 says `1.161` and this says `1.16`, slight but probably important difference.

jschueller:
there is no 1.16.1 version, only 1.16 so #887 can be closed

xhochy:
Thanks!

PR #899, opened 2020-10-27
close a few migrations

ocefpaf:
All these migrations issued all the PRs and are not "standing" migrations" like aarch64, etc.


beckermr (changes requested):
Thanks @ocefpaf! 

I am rethinking our practice of closing these migrators early after all PRs are issued. This is for a few reasons

- If we close them early, we cannot apply bot rerun labels.
- I worry that we lose track of them as a group and packages are getting left behind.
- I think this will make weird solves even harder to diagnose. At least right now, people can look at the status page and see if the thing they cannot get is part of a migration. Otherwise, they will have to check the pinnings and go dig around in feedstocks.

ocefpaf:
@CJ-Wright briefly discussed these at the meeting today. These are mostly "stale" and the failing/unmerged packages are virtually the same across all those migrations. Also, a re-render would bring them up to speed in a single PR instead of multiple bot-reruns. Anyways, no strong opinions, I just want a better looking status page without so many stale migrations to ignore.

beckermr:
can we change the status page instead?

PR #940, opened 2020-11-12
Update zeromq pin to 4.3.3

minrk:
mac-arm builds only have 4.3.3, needed for https://github.com/conda-forge/pyzmq-feedstock/pull/59

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -671,7 +671,7 @@ xrootd:
 xz:
   - 5.2
 zeromq:
-  - 4.3.2
+  - 4.3.3
 zlib:
   - 1.2
 zstd:


xhochy (changes requested):
Shouldn't this require a migration first? (Or is pyzmq the only dependent?)

minrk:
I don't believe so, but I'm not 100% sure. Packages built with 4.3.2 are compatible with 4.3.3 (except on Windows), so this should only affect what is installed at build-time, not compatibility (builds after re-render will be compatible with >=4.3.3 instead of >=4.3.2). So maybe a Windows-only migration? Is that possible?

minrk:
I'll do a migration just to be safe. 4.3.3 isn't the small patch release that it claims to be, so better to do a full migration to catch possible issues.

minrk:
migrator in #941

PR #941, opened 2020-11-12
zeromq 4.3.3 migrator

minrk:
use migrator instead of direct pinning update in #940


beckermr:
cc @conda-forge/zeromq

CJ-Wright:
One thing to note is that I don't think the bot will issue PRs with micro versions since it is behind a selector (and the bot runs on linux usually). If this is something we need to fix long term, we may need to change how the bot renders these and choose the most stringent constraint.

PR #1028, opened 2020-12-18
pin qtkeychain 0.12.0

ocefpaf:
We did not pin this before b/c [we did not need to](https://abi-laboratory.pro/index.php?view=timeline&l=qtkeychain). However, the latest version added a few changes that may be incompatible, see https://github.com/conda-forge/qtkeychain-feedstock/pull/11#issuecomment-747866848

Pinging @gillins.

PS: I guess we'll need a repo data patch too. I never get those right :-/

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -289,6 +289,8 @@ pin_run_as_build:
     max_pin: x.x
   qt:
     max_pin: x.x
+  qtkeychain:
+    max_pin: x.x
   readline:
     max_pin: x
   r-base:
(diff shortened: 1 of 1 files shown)


xhochy:
Shouldn't the pin be 0.11 and 0.12 a migrator (the bot will open that automatically)?

ocefpaf:
> Shouldn't the pin be 0.11 and 0.12 a migrator (the bot will open that automatically)?

I thought about moving directly to the 0.12 and repo patch for 0.11 but your way is probably better. Let's do it!

xhochy (approved):

PR #1138, opened 2021-01-20
Loosen libtiff pinning

xhochy:
`run_exports` is major only, so loosen the pin to allow for different versions https://github.com/conda-forge/libtiff-feedstock/blob/93146b2fc91af119c6649904661777fa739d3308/recipe/meta.yaml#L34-L35

This may unblock some unsolvable migrations.


chrisburr (approved):
Looks good 👍 

https://abi-laboratory.pro/?view=timeline&l=libtiff

PR #1162, opened 2021-01-26
add cuda 11.1 and 11.2

jaimergp:
Checklist


jaimergp:
I am not sure if this is the right way to add more CUDA versions (reusing existing migrators). My rationale is that people that opted-in for CUDA 11.0 will want 11.1 too.

isuruf:
No, reusing wouldn't work as the packages already migrated will get 11.1 builds before their dependencies does.

Can you create a new migrator with 11.1 and 11.2 added?

jaimergp:
Like this?

isuruf:
Do we want to support 7 CUDA versions? This would result in big matrix of builds per feedstock.

isuruf:
Please merge the two migrators. Otherwise you'll get two PRs.

kkraus14:
> Do we want to support 7 CUDA versions? This would result in big matrix of builds per feedstock.

Agreed. Is there a clean way to default to 3 CUDA versions and then allow projects to opt into building additional versions if they want to?

leofang:
> > Do we want to support 7 CUDA versions? This would result in big matrix of builds per feedstock.
> 
> Agreed. Is there a clean way to default to 3 CUDA versions and then allow projects to opt into building additional versions if they want to?

I begin wondering if it's ok to always use the latest 11.x to build for all 11.y with y <= x. I think CUDA 11 provides a strong guarantee in ABI compatibility?

kkraus14:
> I begin wondering if it's ok to always use the latest 11.x to build for all 11.y with y <= x. I think CUDA 11 provides a strong guarantee in ABI compatibility?

nvRTC is the exception to this rule. If we split the `cudatoolkit` feedstock up into separate libraries and update recipes accordingly then we absolutely could for most of the libraries.

jaimergp:
I have been thinking about the split too, possibly mimicking the same scheme Ubuntu follows, but it will require quite some work. I'll open an issue in the cudatoolkit feedstock.

jaimergp:
Is the matrix expansion issue going to be a blocker here or can we address this in the future and merging this one for now?

leofang:
> Is the matrix expansion issue going to be a blocker here or can we address this in the future and merging this one for now?

I believe it's not a blocker. cc: @conda-forge/core

beckermr:
This build matrix is getting very large. Can we at least drop some of the older 10.X versions? People can add them back by inserting an appropriate conda_build_config in their feedstock.

jaimergp:
If that's the case, I think we should then add that procedure to the docs and link to the relevant section from the PR description (if possible).

leofang:
> People can add them back by inserting an appropriate conda_build_config in their feedstock.

Interesting @beckermr, is this already working? What's the effect of not specifying 9.x/10.x in a migration file?

beckermr:
> Interesting @beckermr, is this already working? 

Yes, it works for anything you want to override in the main pinnings file. This is a conda-build feature.

> What's the effect of not specifying 9.x/10.x in a migration file?

It will remove those versions when making migration PRs to feedstocks.

leofang:
> > Interesting @beckermr, is this already working?
> 
> Yes, it works for anything you want to override in the main pinnings file. This is a conda-build feature.

Ah, ok, I knew this already 😂 I thought you were referring to some black magic...

> > What's the effect of not specifying 9.x/10.x in a migration file?
> 
> It will remove those versions when making migration PRs to feedstocks.

Sounds like an interruptive change to me...I know how to manipulate `conda_build_config.yaml` so it's not a problem for me, but I can't speak for others.

leofang:
cc: @jakirkham @kkraus14

kkraus14:
> This build matrix is getting very large. Can we at least drop some of the older 10.X versions? People can add them back by inserting an appropriate conda_build_config in their feedstock.

I would advise to remove 9.2 as well 😄. We may not want the latest version (11.2 as of now) in the build matrix by default as often cloud instances do not have a supported driver for it.

jaimergp:
So, two questions:

* Which versions do we want to add by default? 10.2, 11.0, and 11.1?
* How would one user add more versions to the matrix using `conda_build_config.yaml`? Is it enough to specify `cuda_compiler_version` or do they need to worry about the corresponding zipped keys too (docker_image, cudnn, etc)?

beckermr:
Yes they need to add the zipped keys.

kkraus14:
Another thing to consider is if we only support 11.0+ by default then we can use gcc9 as opposed to needing to do the hacky migrators in order to get gcc7 with cuda 10.x.

jakirkham:
Do we have any visibility into what versions people are using? If not, maybe we should do a poll. This would help inform what is still important to maintain

jaimergp:
Downloads per version:

image

Notebook here: https://colab.research.google.com/drive/1X2ThYvuEZbr3NmvwUNNkHfcSk2y_YnbY?usp=sharing

jaimergp:
* Kaggle [runs on 10.2](https://github.com/Kaggle/docker-python/blob/master/gpu.Dockerfile#L18-L20)
* Colab runs on 10.1 
image

kkraus14:
> * Kaggle [runs on 10.2](https://github.com/Kaggle/docker-python/blob/master/gpu.Dockerfile#L18-L20)
> * Colab runs on 10.1
>   image

Colab only supports Python 3.6 which many projects have stopped supporting, namely Numpy: https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table

Kaggle recently moved to driver version 11.0 (https://www.kaggle.com/titericz/gpu-and-cudf-info?scriptVersionId=53279381) so it shouldn't matter about what version of the toolkit is locally installed in the context of a conda environment.

jakirkham:
Thanks Jaime! 😄 That makes a pretty clear case for dropping 9.2. 10.0 seems like it is on the cusp and 10.1 and 10.2 still see a lot of usage it seems. Not too surprising some of the newer 11 versions haven't caught on yet.

Do you recall if the drivers supported by 10.0, 10.1, and 10.2 differed Keith? 🙂

kkraus14:
> Do you recall if the drivers supported by 10.0, 10.1, and 10.2 differed Keith? 🙂

Yes, there was a new driver with each, but as always the driver is backwards compatible, so they can all run with a sufficiently new driver.

kkraus14:
> Do you recall if the drivers supported by 10.0, 10.1, and 10.2 differed Keith? 🙂

Yes, there was a new driver with each, but as always the driver is backwards compatible, so they can all run with a new driver.

jakirkham:
> > Do you recall if the drivers supported by 10.0, 10.1, and 10.2 differed Keith? 🙂
> 
> Yes, there was a new driver with each, but as always the driver is backwards compatible, so they can all run with a new driver.

Sorry let me rephrase. If we dropped 10.0 and 10.1 only keeping 10.2, is there any hardware that would not be able to run with 10.2? If so, what if we only dropped 10.0 and kept 10.1 and 10.2? IOW how can we maximize the amount of hardware covered with the fewest number 10.x versions supported?

jaimergp:
> That makes a pretty clear case for dropping 9.2

Take into account that 9.2 is not available on Windows, so it has less chances of accumulating downloads.

kkraus14:
> Sorry let me rephrase. If we dropped 10.0 and 10.1 only keeping 10.2, is there any hardware that would not be able to run with 10.2? If so, what if we only dropped 10.0 and kept 10.1 and 10.2? IOW how can we maximize the amount of hardware covered with the fewest number 10.x versions supported?

Newer versions typically continue supporting old hardware architectures unless they're really old. There's a support table here: https://docs.nvidia.com/deploy/cuda-compatibility/index.html#support-hardware

jakirkham:
Maybe we should just start by dropping 9.2 then

jakirkham:
Maybe I should have said, please +1/-1 the comment above ^ if you agree/disagree 🙂

isuruf:
>  If we dropped 10.0 and 10.1 only keeping 10.2, is there any hardware that would not be able to run with 10.2?

Answer to this is question is no right?

jakirkham:
Actually the answer is yes hardware does get dropped even on the minor releases

Given the download counts, it looks like all of those are still seeing a lot of usage. So I'm not sure it is wise to drop them yet.

isuruf:
> Actually the answer is yes hardware does get dropped even on the minor releases (IIUC which I may not have)

Did you read the table that @kkraus14 linked?

jakirkham:
I did, but as I said I may not have read it correctly. If you'd like to make a proposal and back it up with that table, would encourage you to do so 🙂

isuruf:
The table doesn't show any hardware being dropped. Can you show one entry where a newer driver version dropped support for an older hardware?

isuruf:
Since dropping cuda versions doesn't impact hardware supported, I'm suggesting we pick one of the two options below.
1. Build only 11.0, 11.1, 11.2 
2. Build only 9.2, 11.0, 11.1, 11.2 - (If for some reason other than hardware, a user can't update their driver version, then they can use 9.2) This is the same strategy that openblas uses where older hardware support is disabled and falls back to the oldest hardware.

I'm leaning towards 1, but if a conda-forge member objects we can do 2 too.

kkraus14:
> Since dropping cuda versions doesn't impact hardware supported, I'm suggesting we pick one of the two options below.
> 
> 1. Build only 11.0, 11.1, 11.2
> 2. Build only 9.2, 11.0, 11.1, 11.2 - (If for some reason other than hardware, a user can't update their driver version, then they can use 9.2) This is the same strategy that openblas uses where older hardware support is disabled and falls back to the oldest hardware.
> 
> I'm leaning towards 1, but if a conda-forge member objects we can do 2 too.

Your proposal completely skips over a major version of CUDA 10.x, where if someone only supports building with CUDA 10.x they're getting put in an unsolvable situation. How about `9.2`, `10.2`, `11.0`, and `11.1`? I'd suspect most people don't have a driver that supports `11.2` yet anyway.

isuruf:
Okay, added 10.2 to both

jakirkham:
Thanks for spelling that out, Isuru. That sounds reasonable to me 🙂 

Also am ok with 1. There don't seem to be much demand for 9.2 (at least based on download counts). Though admittedly some platforms are not well supported currently on 9.2 as Jaime notes, but I haven't seen anyone put much time in there so expect it won't change in the near future either

WDYT Keith do you have a preference on 1 or 2 given the inclusion of 10.2 on that list?

jaimergp:
So the policy could be worded as:

* Current major version (11), all minor versions except most recent one if it's "too fresh", based on estimated driver support. (Does Anaconda keep some kind of stats of what kind of `__cuda` virtual packages people have? This could be an interesting data source).
* Previous major version (10) latest minor version (10.2).
* Maybe the same for one more previous major version.

kkraus14:
I think the only situation I've seen for libraries wanting to use 9.2 is when they **only** want to use 9.2 because they're typically only using the CUDA Driver API. We should make sure to document how to do this nicely (or split the packages to make it easier to do this since they typically don't even need `nvcc`).

I'm in strong favor of option 1, thanks for writing it up @isuruf. I think we should also make sure to document how to add CUDA 10.1 / 10.0 to a recipe if the maintainers want to do so.

jakirkham:
@jaimergp, would you be able to update this PR to reflect option 1 ( https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/1162#issuecomment-774295979 )? 🙂

jaimergp:
I have [manually updated the migrator](https://github.com/conda-forge/nvcc-feedstock/pull/63/commits/63899dfee71aa1874088ea24249fb48db60af9ef) on https://github.com/conda-forge/nvcc-feedstock/pull/63 and there's a weird issue: [Linux CUDA 10.2 gets the Docker image for 9.2](https://github.com/conda-forge/nvcc-feedstock/pull/63/commits/07f25e5a84f07ed75cc76b56d7a66f3d483d3111#diff-b3a4b9d8e8fbeffd87f01b9f61f4236e8bc764b1d56e34447e0a923f88713ef8). Am I missing something?

h-vetinari:
That sounds like a corner case of the interaction between the global pinning and the migrator (resp. a local `conda_build_config.yaml`), since you're trying to delete an entry from the global pinning with a migrator.

Not sure if this is currently possible, much lesss bug-free?

jaimergp:
Does that mean that, right now, a matrix can only be expanded, but not reduced?

jakirkham:
Maybe because of existing migrators getting lumped together (Windows and CUDA 11.0)?

jaimergp:
I renamed them so they wouldn't get caught by Smithy and didn't make a change, so I guess that's a +1 for the global pinnings conflict hypothesis.

jakirkham:
@CJ-Wright @beckermr, can you please advise us on next steps here?

CJ-Wright:
I think this looks reasonable. I think the next steps are to merge this and baby sit it a little. (Baby sitting being watching the bot's PRs and making certain that the PRs are good an proper for a little while, if something is wrong we revert the PR or pause the migration)

jaimergp:
We do have a [this problem](https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/1162#issuecomment-776652909), though... :/

CJ-Wright:
Hmm ok, I didn't see that. I'm not certain what is causing that, if I had to guess I'd guess it was a zipped keys issue since those are usually difficult to sort out.

jaimergp:
Ok, I guess I'll have to fight Smithy a bit! Any pointers on where to start?

jaimergp:
Thanks @isuruf! What was the problem? `docker_image` and `cuda_compiler_version` need to be as long as originally listed, but they can be trimmed with any other zipped key?

isuruf:
Can you send a PR to remove `9.2, 10.0, 10.1` from the default config in `recipe/conda_build_config.yaml`? (And also to staged-recipes)

h-vetinari:
Seems there's another unexpected interaction: rerendering a win_cuda PR after merging the 11.1/11.2 migrator ends up selecting vs2019 for windows (without other changes to `conda-forge.yml`), see https://github.com/conda-forge/faiss-split-feedstock/pull/19/commits/929890669926da85534b9d39575dd82c146ab205

PR #1187, opened 2021-02-06
add bootstrapping variable to conda_build_config.yaml

ilanschnell:
During the process of bootstrapping packages to a new architecture, it is sometimes necessary to build packages in bootstrapping mode.  For example, when the full functionality of the package isn't available yet, but the package is necessary to build other packages will make the full functionality available later (in a regular non-bootstrapping) build.

In order to decide whether or not a build is run in bootstrapping mode, I suggest adding a new variable to `conda_build_config.yaml` named `bootstrapping` whose value is either "yes" or "no".

In a recipe, we can then distinguish packages created by using:
```
build:
{% if bootstrapping == 'yes' %}
  number: 0
  string: h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}_bootstrapping
{% else %}
  number: 1
{% endif %}
```
For example, we can disable build requirements the following way:
```
requirements:
  build:
{% if bootstrapping != 'yes' %}
    - libtool  # [unix]
    - bash     # [unix]
{% endif %}
```
In some cases it may the necessary to have access to the bootstrapping variable from within `build.sh`:
```
  script_env:
    - BOOTSTRAPPING={{ bootstrapping }}
```

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -161,6 +161,11 @@ BUILD: aarch64-conda_cos7-linux-gnu     # [aarch64]
 cdt_arch: armv7l                          # [armv7l]
 BUILD: armv7-conda_cos7-linux-gnueabihf   # [armv7l]
 
+# bootstrapping specifies whether or not a recipe is run in "bootstrapping"
+# mode during the initial progress of adding a new platform.  Most recipes
+# will not need this value.  Possible values are: "yes" and "no".
+bootstrapping: no
+
 # TODO: remove these when run_exports are added to the packages.
 pin_run_as_build:
   arpack:


mingwandroid:
LGTM

ilanschnell:
It looks like https://github.com/conda-forge/patch-feedstock/pull/8/files already relies on `bootstrapping` to be set in `conda_build_config.yaml`.

jakirkham:
As requested in the `make` PR ( https://github.com/conda-forge/make-feedstock/pull/28 ), please raise an issue on the webpage repo for discussion. Thanks!

https://github.com/conda-forge/conda-forge.github.io

cc @conda-forge/core

CJ-Wright:
Are there docs for the bootstrapping variable, the bootstrapping process and how this should be used?

mingwandroid:
It's not formally baked into any tools such as conda or conda-build. It relies simply on eliding dependencies (and relying on the system to provide them) by setting a jinja2 variable called 'bootstrapping'.

When this is done, the package build string is modified to include it so that bootstrapping packages are obvious and can be skipped by tooling. Also, for these packages, we keep to bootstrapping version's build number at 0, so that non-bootstrapping are preferred.

I'd like to get agreement on the 'how' before I document it. Also, I'm not sure where it would be best documented, suggestions welcome.

hmaarrfk:
Discussion does not seem to have been resolved. 
https://github.com/conda-forge/conda-forge.github.io/discussions/1255

Marking this as draft until it is.

PR #1190, opened 2021-02-08
add gdal to the osx-arm64

ocefpaf:
Because why not? Let's see how far we already have and how bad it will break.


ocefpaf:
That was fast! Thanks!!

PR #1351, opened 2021-03-18
Added jax to osx_arm64.txt

storopoli:
Added Google's JAX to osx_arm64.txt. It is a very important and used dependency (e.g. NumPyro depends on it)


xhochy:
We already have this in the migration graph. We need a from-source build in https://github.com/conda-forge/jaxlib-feedstock/pull/51 / https://github.com/conda-forge/jaxlib-feedstock/pull/49 to get this working.

PR #1714, opened 2021-07-21
unpause win-pypy migration

h-vetinari:
Follows #1668 & #1670, now that numpy has been built, see [here](https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/1668#issuecomment-884125276).

CC @isuruf @beckermr @mattip

### DIFF
--- a/recipe/migrations/pypy37-windows.yaml
+++ b/recipe/migrations/pypy37-windows.yaml
@@ -1,6 +1,6 @@
 migrator_ts: 1623865877
 __migrator:
-    paused: True
+    paused: False
     migration_number: 1
     operation: key_add
     primary_key: python


mattip:
Do you understand why the [migration status](https://conda-forge.org/status/#pypy37-windows) shows matplotlib as a child of pyqt (in the In-PR section)? The [recipe](https://github.com/conda-forge/matplotlib-feedstock/blob/1fb8f9baf8b74041daf9a84a115d00653c9e1cab/recipe/meta.yaml#L83) only requires pyqt for CPython

h-vetinari:
I think that's a simple case of the migrator not taking into account selectors. In such cases (barring changes to the migrator, cf. https://github.com/regro/cf-scripts/issues/1331 / https://github.com/regro/cf-scripts/issues/1332), you can simply do a manual migration like https://github.com/conda-forge/numpy-feedstock/pull/238

mattip:
xref conda-forge/matplotlib-feedstock#298. Once (if?) that goes in, will the migrator then notice and continue from there?

h-vetinari:
> Once (if?) that goes in, will the migrator then notice and continue from there?

No, the migrator only knows its own PRs. But one could then close the pyqt one, and would then continue, I believe.

PR #1873, opened 2021-09-13
Add `neovim` to macOS ARM migration

benjaminrwilson:


jakirkham:
Thanks Benjamin! 😀

In the future can you please fill out the title like above? It makes it easier for us to tell what different PRs are doing. Thanks! 🙏

PR #2004, opened 2021-10-12
add python-build

ocefpaf:
Closes https://github.com/conda-forge/python-build-feedstock/issues/13


jakirkham:
Thanks Filipe! 😄 Should we add it to the arch migration as well?

PR #2065, opened 2021-10-26
Close out migration for spdlog190

ryanvolz:
Migration is actually finished. The bot has `librmm` as "not solvable", but the migration for that has been done manually: https://github.com/conda-forge/librmm-feedstock/pull/13.


wolfv:
@ryanvolz looks good to me. I agree with your comment on using `1.9` instead of `1.8`.

PR #2180, opened 2021-11-13
add C/C++ libraries antic & e-antic

saraedum:


recipe/conda_build_config.yaml
```diff
@@ -487,6 +489,8 @@ libcrc32c:
   - 1.1
 libdap4:
   - 3.20.6
+libeantic:
+  - 1.0.3
```
xhochy, line 493: Given the `run_exports` this should be simplified: https://github.com/conda-forge/e-antic-feedstock/blob/4729f28f3d73ced0a6bbb63929356337021cdc41/recipe/meta.yaml#L22
```suggestion
  - 1
```

saraedum: Thanks, @xhochy, you are correct of course.

recipe/conda_build_config.yaml
```diff
@@ -323,6 +323,8 @@ abseil_cpp:
   - '20210324.2'
 alsa_lib:
   - 1.2.3
+antic:
+  - 0.2.4
```
xhochy, line 327: Given https://github.com/conda-forge/antic-feedstock/blob/93dc1d2f492634a1169c8b4e97410567823ce96e/recipe/meta.yaml#L17 this should also be changed
```suggestion
  - 0.2
```

saraedum: Sorry, I thought that we had an exact pin for antic. But you are right.

```suggestion
  - 0.2
```

PR #2202, opened 2021-11-18
Add superlu_dist

jan-janssen:
In addition to https://github.com/conda-forge/superlu_dist-feedstock/pull/35

Checklist

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -692,6 +692,8 @@ sqlite:
   - 3
 suitesparse:
   - 5
+superlu_dist:
+  - 7
 tbb:
   - 2020
 tbb_devel:
--- a/recipe/migrations/superlu_dist711.yaml
+++ b/recipe/migrations/superlu_dist711.yaml
@@ -1,0 +1,8 @@
+migrator_ts: 1631819618
+__migrator:
+  kind: version
+  migration_number: 1
+  bump_number: 1
+
+superlu_dist:
+  - 7.1.1


xhochy:
As this is not a totally new version, should we also add a migration to move all packages to the same pin?

jan-janssen:
@xhochy This is the first time I define the pinning, so I am not exactly sure. My original issue was: `fenics` depends on `petsc` and `superlu_dist` and `petsc` itself depends on `superlu_dist`. When `superlu_dist` is updated and a new build of `petsc` is triggered it increases the build number and links to the new `superlu_dist` version so until `fenics` is updated all installations fail as the users install `fenics` with the updated `petsc` version but the old `superlu_dist` version. To my understanding this could be fixed by adding the `run_export` to `superlu_dist`. https://github.com/conda-forge/superlu_dist-feedstock/pull/35 and this pull request - Is this correct?

xhochy:
> @xhochy This is the first time I define the pinning, so I am not exactly sure. My original issue was: `fenics` depends on `petsc` and `superlu_dist` and `petsc` itself depends on `superlu_dist`. When `superlu_dist` is updated and a new build of `petsc` is triggered it increases the build number and links to the new `superlu_dist` version so until `fenics` is updated all installations fail as the users install `fenics` with the updated `petsc` version but the old `superlu_dist` version. To my understanding this could be fixed by adding the `run_export` to `superlu_dist`. [conda-forge/superlu_dist-feedstock#35](https://github.com/conda-forge/superlu_dist-feedstock/pull/35) and this pull request - Is this correct?

Yes, this is exactly what this will achieve. However, also adding a migration at the same time will ensure that the bot issues PRs to all packages depending on `superlu_dist` to ensure that they are rebuilt with the correct pinning. As you are adding a `run_exports`, I would suggest to also do this here. 

At the same time, you probably should also write a repodata patch that retrospectively adds a stricter `superlu_dist` pinning to the existing `petsc` and `fenics` builds.

jan-janssen:
Ok, I understand that I need to add a migrator file: 
```
migrator_ts: 1631819618
__migrator:
  kind: version
  migration_number: 1
  bump_number: 1

superlu_dist:
  - 7.1.1
```
But I am not exactly sure about the `migrator_ts` variable. In terms of the retrospective patches, I guess we should be fine there are only a couple of packages which need to be updated `proteus`, `precice`, `petsc` and `cashocs`. From these  `proteus`, `precice` and `petsc` are already updated and I just opened a pull request for `cashocs` https://github.com/conda-forge/cashocs-feedstock/pull/16 . So we should be save for now, I just want the migration to be save in the future.

xhochy:
> But I am not exactly sure about the migrator_ts variable. 

This is fine, we only need to be careful about it when we change it afterwards. But this isn't done in 99% of all cases.

jan-janssen:
@xhochy I added the migrator yaml file.

PR #2314, opened 2021-12-15
Add pinning for libglib

hmaarrfk:
I don't really think we need a migrator for this.

But maybe I can add a migrator to ask people to rename their dependency from `glib` to `libglib` if they don't need python.

Checklist

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -521,6 +521,8 @@ libgdal:
   - '3.3'
 libgit2:
   - '1.3'
+libglib:
+  - '2'
 libiconv:
   - 1.16
 libkml:


hmaarrfk:
I guess i can't figure out how to create a migrator that attempts to do `glib` -> `libglib` in the same way that `matplotlib` -> `matplotlib-base` happened.

If somebody could help with this that would be great.

xhochy:
People should omit `glib` from `run`; having it in `host` adds the correct `libglib` run_export.

hmaarrfk:
got it

hmaarrfk:
@xhochy in the recipe for Qt (5) depenending on `glib` pulls in `python` in turn which pulls in other dependencies that are quite heavy. I understand that Qt is rather hard to build, and we might want to omit things that are "specifiacally for Qt" here, but do you think there are other cases where depending on `libglib` in `host` is better than depending on `glib` in `host`? If so having this pinning would be good for that.

Thank you for your time reviewing this.

xhochy:
Happy to merge this but I don't understand whether this helps you at all.

hmaarrfk:
My local build is nearly done. I'll report soon.

hmaarrfk:
it might be that the loads of other dependencies in Qt make it enough...

hmaarrfk:
Well both Qt and dbus were able to compile successfully with just `libglib` however, I think you are much more in tune with the "correct" approach. I was mostly hacking to get to the bottom of the build challenges in Qt.

https://github.com/conda-forge/dbus-feedstock/pull/23   <-- See 2nd commit where `libglib` was used instead of `glib`. I reverted to the use of `glib` in `host` (nothing in `run`) as per your suggestion.

xhochy:
@hmaarrfk Can we close this or is there a valid reason for the PR?

hmaarrfk:
sorry for letting this linger here

PR #2537, opened 2022-02-21
Add pins for trelliscopejs dependencies

colinbrislawn:


hmaarrfk:
Why not just add `trelliscopejs`

colinbrislawn:
`r-trelliscopejs` is not yet on conda-fordge, but hey, it would awesome if it was! It builds just fine on the M1 mac

hmaarrfk:
got it.

PR #2628, opened 2022-03-19
fortls migration

awvwgk:
https://github.com/conda-forge/fortls-feedstock


hmaarrfk:
As a tip, maybe don't add to the bottom of the list. It increases the chance of conflicts.

PR #2690, opened 2022-03-31
Fix CUDA ppc64le/aarch64 migrator (again)

leofang:
Follow up of #2687. 

cc: @jakirkham @isuruf @kkraus14 

Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #2695, opened 2022-04-01
Fix CUDA ppc64le/aarch64 migrator (final try)

leofang:
Follow-up of #2690. 

Verified locally (by intercepting the global pinning inside conda-smithy and rewriting it). This is needed to make `conda_build.api.render()` know the existence of `cuda_compiler` on the supported platforms. Otherwise it'd mistakenly generate entries like `cuda_linux-aarch64`, which should not happen as back in the old days overwriting with a local cbc.yaml worked. 

cc: @isuruf @jakirkham @kkraus14 

Checklist

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -50,8 +50,8 @@ m2w64_fortran_compiler:        # [win]
 CMAKE_GENERATOR:               # [win]
   - NMake Makefiles            # [win]
 
-cuda_compiler:                 # [linux64 or win or aarch64 or ppc64le]
-  - nvcc                       # [linux64 or win or aarch64 or ppc64le]
+cuda_compiler:                 # [linux or win]
+  - nvcc                       # [linux or win]
 cuda_compiler_version:
   - None
   - 10.2                       # [(linux64 or win) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -50,8 +50,8 @@ m2w64_fortran_compiler:        # [win]
 CMAKE_GENERATOR:               # [win]
   - NMake Makefiles            # [win]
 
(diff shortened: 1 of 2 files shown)


recipe/conda_build_config.yaml
```diff
@@ -50,8 +50,8 @@ m2w64_fortran_compiler:        # [win]
 CMAKE_GENERATOR:               # [win]
   - NMake Makefiles            # [win]
 
-cuda_compiler:                 # [linux64 or win]
-  - nvcc                       # [linux64 or win]
+cuda_compiler:                 # [linux64 or win or aarch64 or ppc64le]
+  - nvcc                       # [linux64 or win or aarch64 or ppc64le]
```
jakirkham, line 54:
```suggestion
cuda_compiler:                 # [linux or win]
  - nvcc                       # [linux or win]
```

leofang: Thx. I thought about this but was worried it's not general enough.

leofang:
Do we need to re-trigger the migrator run? Apparently the old runs were not valid.

recipe/conda_build_config.yaml
jakirkham: Yeah `linux` covers all 3 architectures

jakirkham:
Yeah that makes sense. Would send a new PR to bump [this number]( https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/9d93e4cc5491c5a53df68e88cea6ee1febb0333d/recipe/migrations/cuda_112_ppc64le_aarch64.yaml#L5-L6 ) with an explanation of why it is needed.

jakirkham:
Thanks for all the hard work here Leo! 😄

jakirkham:
Just a note, this is a no-op on a non-migrated feedstock (as expected). Tested in PR ( https://github.com/conda-forge/ucx-split-feedstock/pull/101 )

PR #2707, opened 2022-04-04
Update osx_arm64.txt

mlund:
Request for arm64 target for faunus-feedstock


jakirkham:
@mlund, would suggest writing the PR title to include the feedstock name that is being added in the future. Something like "Adding fanus to macOS ARM migration" or similar

PR #2726, opened 2022-04-06
ENH close proj810

beckermr:
The PR that closed this one was accidentally closed by me and so this one is still active even though it is done. Removing it by hand now.


beckermr:
No pinning chaneg is needed since proj is at 8.2.X now.

PR #2727, opened 2022-04-06
ENH remove libunwind16

beckermr:
This PR closes the libunwind16 migration. It is was added by hand by @hmaarrfk and has now issued all of its PRs. The main pinnings do not need an update since they are at 1.6 already.


beckermr:
@hmaarrfk Can you confirm this is ok to merge?

hmaarrfk:
Thanks

PR #2860, opened 2022-05-11
Let lz4-c manage its own pinnings

hmaarrfk:
It seems that they only want to to pin to `x.x` not `x.x.x` 
https://github.com/conda-forge/lz4-c-feedstock/blob/main/recipe/meta.yaml#L37

Checklist

### DIFF
--- a/recipe/migrations/lz4_c19.yaml
+++ b/recipe/migrations/lz4_c19.yaml
@@ -1,0 +1,8 @@
+migrator_ts: 1652241058
+__migrator:
+  kind: version
+  migration_number: 1
+  bump_number: 1
+
+lz4_c:
+  - '1.9'
--- a/recipe/migrations/lz4_c19.yaml
+++ b/recipe/migrations/lz4_c19.yaml
@@ -5,4 +5,4 @@ __migrator:
   bump_number: 1
 
 lz4_c:
-  - '1.9'
+  - '1.9.3'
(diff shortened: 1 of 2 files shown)


hmaarrfk:
We probably need a patch to loosen the lz4-c pinning after this goes through.

jakirkham:
Should we change this to `1.9`?

https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/9dbffc09022341fb0ed0a76053639c88e94e55cd/recipe/conda_build_config.yaml#L606-L607

Think we want to capture this in [a migrator]( https://github.com/conda-forge/cfep/blob/main/cfep-09.md#1-simple-additive-version-migrations )

hmaarrfk:
you are right. let me change that

hmaarrfk:
Let me know what you think.

jakirkham:
Seems ok to me

@conda-forge/core, any other thoughts here? 🙂

hmaarrfk:
can we have a pinning that has an inequality constraint? Like `>1.9.3,<1.10.0a`? I think that would be "more safe".

recipe/migrations/lz4_c19.yaml
```diff
@@ -0,0 +1,8 @@
+migrator_ts: 1652241058
+__migrator:
+  kind: version
+  migration_number: 1
+  bump_number: 1
+
+lz4_c:
+  - '1.9'
```
jakirkham, line 8: We could set this to...

```suggestion
  - '1.9.3'
```

...for now and then drop the patch version when we move to `1.10` or `2` whichever is next

jakirkham:
Thanks Mark! 🙏 Let's give this a try

PR #2886, opened 2022-05-18
Migrate to petsc 3.17

minrk:
<del>The petsc 3.16 migration still has two outstanding packages (daetk, proteus), [stuck on daetk](https://github.com/conda-forge/daetk-feedstock/pull/6). Should I complete the 3.16 migration here too, or let it keep going?</del> 3.16 migration is complete

closes #2732 
closes #2744


minrk:
3.16 migration has completed, so we can safely start with 3.17

PR #2973, opened 2022-06-07
Pin p11-kit at 0.23.21

leofang:
Checklist

### DIFF
--- a/recipe/migrations/p11kit023.yaml
+++ b/recipe/migrations/p11kit023.yaml
@@ -1,0 +1,8 @@
+migrator_ts: 1654633916
+__migrator:
+  kind: version
+  migration_number: 1
+  bump_number: 1
+
+p11_kit:
+  - 0.23.21


leofang:

hmaarrfk:
This stuff always hurts my head. Because now you need to babysit every build to ensure that it picked up the right build number.

I'm not sure if we should mark the older builds as broken.

leofang:
I am not sure I follow, Mark. My understanding is once this migrator is merged, the bot will rebuilt all packages that depend on `p11-kit`, and then once the rebuild is done we're good. Why do we need to mark all old packages broken? I haven't seen such things done in the past.

hmaarrfk:
you added the run_exports to the latest build. There is no guarantee the solver will find it. It may take a few hours for it to propagate. 

So you might get build `3` which doesn't have the run exports, and sometimes you will get build `4` which does.

leofang:
We can't be always "past compatible" right? This is to ensure we fix things properly to the future and do not look back unless there are serious issues.

I am doing this PR mainly with the same concern in mind. The bot may issue p11 0.24.0 PR anytime, and I want to get the pinning done before the new version is merged. Otherwise, once the new version is out and we still don't have the right pinning, things get worse (than they are now, as 0.23 has been around for years.

hmaarrfk:
ok. lets rebuild. but you might also want to add a repo-data patch to add the max pin to 0.23 for previous builds.

leofang:
Will do, thanks Mark, though I'd likely need your help for the repodata patch 😉

PR #3045, opened 2022-06-25
relax abseil_cpp and grpc_cpp?

ngam:
Tensorflow can only be built with the latest version anyway, so we might as well move forward with these pins now? I would defer to isuruf, xhochy, and hmaarrfk on how to best move forward here. But my point is that we don't need to build 20210324.2 anymore.

Checklist

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -169,9 +169,6 @@ zip_keys:
     - python
     - numpy
     - python_impl
-  -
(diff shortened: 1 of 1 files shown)


beckermr (changes requested):
We use migrations for these pinning changes.

beckermr:
The migration would handle any other packages that depend on these pins. Sure for tensorflow we don't need it technically.

PR #3175, opened 2022-07-28
Update osx_arm64.txt with rubin_sim

jmeyers314:
Checklist

### DIFF
--- a/recipe/migrations/osx_arm64.txt
+++ b/recipe/migrations/osx_arm64.txt
@@ -959,3 +959,4 @@ iteration_utilities
 hmmlearn
 arm_pyart
 stockwell
+rubin_sim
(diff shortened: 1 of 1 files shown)


recipe/migrations/osx_arm64.txt
```diff
@@ -959,3 +959,4 @@ iteration_utilities
 hmmlearn
 arm_pyart
 stockwell
+rubin_sim
```
beckermr, line 962:
```suggestion
rubin-sim
```

beckermr: I think it needs to be this?

jmeyers314: I was following along with https://github.com/lsst/rubin_sim#conda-installation

beckermr: Right. This is the name of the feedstock, not the package iirc

PR #3159, opened 2022-07-26
Roll back abseil migration

hmaarrfk:
Given the concerns @h-vetinari raised and the challenges in 
https://github.com/conda-forge/grpc-cpp-feedstock/pull/193

I think it is best to compile grpc first, then to start a migration when that package is ready.

Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #3181, opened 2022-07-29
Abseil 20220623 migration (redux)

h-vetinari:
Follow-up to #3159, using the builds added in https://github.com/conda-forge/abseil-cpp-feedstock/pull/35. gRPC is passing with those builds in https://github.com/conda-forge/grpc-cpp-feedstock/pull/196. 

~This is obviously blocked on the abseil PR, but~ I wanted to see what changes would be necessary to take care of the `abseil_cpp` -> `libabseil` rename.

Closes #3220

CC @hmaarrfk @mariusvniekerk @xhochy

### DIFF
--- a/recipe/migrations/abseil_cpp202206230.yaml
+++ b/recipe/migrations/abseil_cpp202206230.yaml
@@ -1,0 +1,13 @@
+__migrator:
+  build_number: 1
+  kind: version
+  migration_number: 1
+
+# TODO: Make sure to remove the associated zip keys in the final file
+# when the migration is complete
+abseil_cpp:
+- '20220623.0'
+grpc_cpp:
+- '1.48'
+
+migrator_ts: 1656693919.0942996
--- a/recipe/migrations/abseil_cpp202206230.yaml
+++ b/recipe/migrations/abseil_cpp202206230.yaml
@@ -15,6 +15,6 @@ libabseil:
 libabseil_static:
 - '20220623.0'
 grpc_cpp:
-- '1.48'
+- '1.46'
 
 migrator_ts: 1656693919.0942996
--- a/recipe/migrations/abseil_cpp202206230.yaml
+++ b/recipe/migrations/abseil_cpp202206230.yaml
@@ -7,6 +7,13 @@ __migrator:
 # when the migration is complete
 abseil_cpp:
 - '20220623.0'
+# new name, use both for now
+libabseil:
+- '20220623.0'
+# separate output for static builds (depending on C++ version)
+# also need to be migrated
+libabseil_static:
+- '20220623.0'
 grpc_cpp:
 - '1.48'
 
--- a/recipe/migrations/abseil_cpp202206230.yaml
+++ b/recipe/migrations/abseil_cpp202206230.yaml
@@ -1,7 +1,7 @@
 __migrator:
   build_number: 1
   kind: version
-  migration_number: 1
+  migration_number: 2
 
 # TODO: Make sure to remove the associated zip keys in the final file
 # when the migration is complete
--- a/recipe/migrations/abseil_cpp_removal.yaml
+++ b/recipe/migrations/abseil_cpp_removal.yaml
@@ -1,0 +1,13 @@
+__migrator:
+  bump_number: 0
+  kind: deletion
+  migration_number: 1
+  wait_for_migrators:
+    - abseil_cpp202206230
+
+# TODO: Make sure to remove the associated zip keys in the final file
+# when the migration is complete
+abseil_cpp:
+- '20220623.0'
+
+migrator_ts: 1660061100


recipe/migrations/abseil_cpp202206230.yaml
```diff
@@ -0,0 +1,20 @@
+__migrator:
+  build_number: 1
+  kind: version
+  migration_number: 1
```
hmaarrfk, line 4:
```suggestion
  migration_number: 2
```

h-vetinari:
@beckermr @chrisburr @hmaarrfk 
Is there a way for a migration to rename a package (i.e. remove the old one, not just add a new one)?

chrisburr:
The CFEP is the best source of docs I know about: https://github.com/conda-forge/cfep/blob/2409812d68296fcf2e06255687cc7cb085923ffd/cfep-09.md#3-name-migrations

h-vetinari:
> The CFEP is the best source of docs I know about

Thanks a lot! Sounds like the best way would be to migrate with just additional keys at first (because we're not changing the content of an existing pin - like `c_compiler` in the example - but renaming the pin key itself), and then eventually run another migration that removes `abseil-cpp`.

recipe/migrations/abseil_cpp202206230.yaml
```diff
@@ -0,0 +1,20 @@
...
+# TODO: Make sure to remove the associated zip keys in the final file
+# when the migration is complete
+abseil_cpp:
+- '20220623.0'
+# new name, use both for now
+libabseil:
+- '20220623.0'
+# separate output for static builds (depending on C++ version)
+# also need to be migrated
+libabseil-static:
```
h-vetinari, line 15:
```suggestion
libabseil_static:
```

Fails with hyphen...

h-vetinari:
@hmaarrfk @mariusvniekerk @ngam

There are now three green gRPC buils against the new abseil version:
* 1.48: https://github.com/conda-forge/grpc-cpp-feedstock/pull/196
* 1.47: https://github.com/conda-forge/grpc-cpp-feedstock/pull/195
* 1.46: https://github.com/conda-forge/grpc-cpp-feedstock/pull/194

Would be nice if you could take a look there & in this PR. :)

PS. I attached a follow-up migrator that removes `abseil_cpp` once the pin has changed to the new `libabseil`.

h-vetinari:
This supersedes #3220

recipe/migrations/abseil_cpp202206230.yaml
```diff
@@ -0,0 +1,20 @@
...
+libabseil:
+- '20220623.0'
+# separate output for static builds (depending on C++ version)
+# also need to be migrated
+libabseil_static:
+- '20220623.0'
+grpc_cpp:
+- '1.48'
+
+migrator_ts: 1656693919.0942996
```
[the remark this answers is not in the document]
h-vetinari: Yes, I used it explicitly in https://github.com/conda-forge/grpc-cpp-feedstock/pull/196 (with the only exception that I had to also do `use_local: True`); it would also be possible to do this for the other PRs, but there @hmaarrfk had already started modifying the CBC, so I stayed with that. Happy to use the migrator there as well if that's desired.

h-vetinari:
@conda-forge/abseil-cpp @conda-forge/grpc-cpp 
PTAL

recipe/migrations/abseil_cpp202206230.yaml
```diff
@@ -0,0 +1,20 @@
...
+- '20220623.0'
+# new name, use both for now
+libabseil:
+- '20220623.0'
+# separate output for static builds (depending on C++ version)
+# also need to be migrated
+libabseil_static:
+- '20220623.0'
+grpc_cpp:
+- '1.48'
```
h-vetinari, line 18: grpc builds have been merged. I don't see another reason for the zip_keys between abseil & grpc, except historical difficulty & convenience.

Still, obviously 1.46-1.48 work fine with the new abseil, so I think it'd be good to take out the zip_keys both from the migratior and the global pinning. That way, feedstocks can choose their own speed with updating grpc (not least since 1.47+ now requires at least C++14, and this might be necessary to propagate to consuming feedstocks)

h-vetinari:
> so I think it'd be good to take out the zip_keys both from the migratior and the global pinning. 

Aside from the possible improvement w.r.t to the coupling of grpc & abseil, this PR remains ready

h-vetinari:
PS. Just saw that the grpc feedstock still has ABI migration branches for 1.43-1.45. I could also backport the new abseil builds there, if needed.

h-vetinari:
> @chrisburr: The CFEP is the best source of docs I know about: https://github.com/conda-forge/cfep/blob/2409812d68296fcf2e06255687cc7cb085923ffd/cfep-09.md#3-name-migrations

I had a look at the cfep again, but AFAICT the ability to add a custom message is only implemented for special migrators like the arch rebuild? I'd like to add a message to the bot PRs along the following lines (but this shouldn't hold up this PR, I can also monitor the bot PRs manually):

> The main package has been renamed from `abseil-cpp` to `libabseil`, also due to some infrastructural changes - please update this in your `meta.yaml`. This is because the ABI of `libabseil` depends on the version of the C++ standard used to compile it, and crucially, this needs to match the C++ standard versions used to compile consuming packages like this feedstock. If your feedstock is not yet ready to compile with C++17 (the version used to compile the shared `libabseil` builds), you can fall back to a static builds for the right C++ version, e.g. `libabseil-static =*=cxx14*` for C++14. Don't forget to add a second unpinned line of `libabseil-static` so future migrators will update the pin nevertheless. Also, if you use the static builds, be aware that you need to bundle the abseil license (Apache 2) into the feedstock. If you have questions, please reach out to @conda-forge/abseil-cpp.

recipe/migrations/abseil_cpp202206230.yaml
hmaarrfk: There are some peculiarity with migrating. I'm worried that without the zip, something will fail. We can take it out once it has all been merged. IMO

h-vetinari: I don't feel very strongly about this (so fine either way) but not sure it helps TBH. If we merge the end of this migration we'll be at 1.48 globally, and downgrading is even more of a hassle.

Any feedstock with problems could still try any grpc between 1.46 - 1.48.

But maybe the golden middle is just to zip this with 1.46, so we're not forcing all feedstocks to make (as big a jump) in grpc:
```suggestion
- '1.46'
```

h-vetinari:
> PS. Just saw that the grpc feedstock still has ABI migration branches for 1.43-1.45. I could also backport the new abseil builds there, if needed.

I saw that grpc is already at 1.45 in the global pinning for the oldest abseil, and 1.46 for 20211102. So the old ABI migration branches below 1.45 aren't relevant anymore, and 1.45 will be after this migration.

I'd still like to zip the newest abseil with 1.46 TBH, as proposed in the review comments above. Afterwards we take out the zip & try a migration for 1.47 or 1.48.

PS. I rebuilt the new abseil outputs also for 20211102 so that the migration PRs don't have to skip the second abseil version that still comes from the global pinning.

h-vetinari:
@conda-forge/core
Any remaining concerns here? Can we kick this off?

hmaarrfk:
so specifically, this would move us to a newer abseil, whlie leaving grpc unchanged. Finally, we would have both outputs decoupled.

Is this a good summary?

h-vetinari:
> Finally, we would have both outputs decoupled.
> 
> Is this a good summary?

Yes.

And then continue to move grpc separately

hmaarrfk:
Let's try

PR #3326, opened 2022-08-25
Ignore pyqt for matplotlib in pypy38

chrisburr:
Fixes the issue mentioned in https://github.com/conda-forge/matplotlib-feedstock/pull/322#issuecomment-1200449398 using the same method that was previously used in https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/1821.

### DIFF
--- a/recipe/migrations/pypy38.yaml
+++ b/recipe/migrations/pypy38.yaml
@@ -27,6 +27,9 @@ __migrator:
       - python
       - pypy3.6
       - pypy-meta
+    ignored_deps_per_node:
+      matplotlib:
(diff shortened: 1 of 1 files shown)


mattip:
Thanks, it seems the migration is unstuck

mattip:
I think this should also be done for jupyter_core. On the migration status, jupyter_core appears as a child of of pywin32 (in the `in-pr` list), but pywin32 does not appear in the jupyter_core parents list (in `awaiting-parents`). Where should these exceptions be documented for the next migration?

PR #3399, opened 2022-09-10
close out migrations for abseil202206230 & grpc_cpp147

h-vetinari:
I've worked through all the issues now, and with the exception of arrow 6.0.x (the oldest of the 4 abi branches for arrow), everything could move to C++17, despite the concerns discussed in https://github.com/conda-forge/abseil-cpp-feedstock/issues/45.

Quoting from the recap there:
> The last open PRs are [tensorflow](https://github.com/conda-forge/tensorflow-feedstock/pull/266) (not surprising), one arrow-cpp migration [branch](https://github.com/conda-forge/arrow-cpp-feedstock/pull/830), and [or-tools](https://github.com/conda-forge/or-tools-feedstock/pull/24). The last one is concerning me a bit, because it doesn't show up in the migrator [status](https://conda-forge.org/status/#abseil_cpp202206230), so I don't know how many other packages I might still be missing, but the number should hopefully be low.
> 
> I'm thinking we could do one more abseil migration for `arrow-cpp` with https://github.com/conda-forge/abseil-cpp-feedstock/pull/48, but switch [the compatibility wrapper] `arrow-cpp` to be the shared C++17 build on windows. That should flush out any remaining stragglers. Then I can fix those, and for the next abseil version, we can remove the compat output.

Crucially, there's now no reason why grpc & abseil should depend on each other, so remove the zip. I've replaced the zip with the other libabseil-builds. It's debatable if we want to keep the `libabseil_static` builds (now that everything seems to be C++17 compatible), but while we do have them, add them to the pinning.

The libprotobuf migration is still ongoing, but all packages that are _also_ affected by abseil/grpc are fine (with two still unmerged PRs as noted above).

CC @conda-forge/abseil-cpp @conda-forge/grpc-cpp @hmaarrfk

PS. I've also removed the `abseil_cpp_removal` migrator I intended in #3181, because it is non-functional AFAICT, and I don't have enough [information](https://github.com/conda-forge/conda-smithy/issues/1674) to fix it.


hmaarrfk:
Let's wait for mongodb

h-vetinari:
I looked at the [code search](https://cs.github.com/conda-forge?q=abseil-cpp) again, and found some mode feedstocks.
- https://www.github.com/conda-forge/cartographer-feedstock
- https://www.github.com/conda-forge/obake-feedstock

I can tackle those before or after merging this, it doesn't really matter (the or-tools PR is also not merged yet)...

h-vetinari:
> I can tackle those before or after merging this, it doesn't really matter IMO

On second thought, let's first try switching those and https://github.com/conda-forge/mongodb-feedstock/pull/60 to C++17 on windows as well.

hmaarrfk:
Thank you for the reviews. I made things a checklist :D

h-vetinari:
> Thank you for the reviews. I made things a checklist :D

I added the or-tools PR to the checklist. :)

h-vetinari:
> * (but maybe we can merge before this one goes through)

Seeing that you added tensorflow to the checklist, that definitely moves on its own timescale and shouldn't hold up things here IMO (also, that PR showed that things are buildable already).

hmaarrfk:
> Seeing that you added tensorflow to the checklist, that definitely moves on its own timescale and shouldn't hold up things here IMO (also, that PR showed that things are buildable already).

it was more just calling it out.

h-vetinari:
I propose not to wait for or-tools. The build runs through and the concerns on that PR are completely unrelated to abseil, and will have to get solved for the next PR in any case (independently of where the pinning stands). Mongodb also ran through now. :)

h-vetinari:
@hmaarrfk 
Merging this would help me a bunch with trying to fix upstream arrow CI (which tests the conda builds, and needs some sync with the yamls we generate; merging this would help get rid of a bunch of unnecessary divergence).

Admittedly, that's not very urgent, but OTOH I don't think we have much to gain from waiting much more here (all open PRs are in a passing state).

h-vetinari:
PR for arrow-cpp 6.0 got unblocked 🥳

h-vetinari:
Ready now @hmaarrfk?

xhochy (approved):
🎉

kkraus14 (approved):

h-vetinari:
Can we please merge this? 🙃 
I need this for some work on arrow.

hmaarrfk:
I think it is ok. or-tools and tensorflow are leaf packages and seem to be going through a final review process.

hmaarrfk:
Thanks for the great work!

PR #3418, opened 2022-09-12
Add C++ library eclib

saraedum:
since eclib does not follow any semantic versioning scheme, we have to pin exactly. That's also what the `run_exports` of eclib does.

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -424,6 +424,8 @@ davix:
   - '0.8'
 dbus:
   - 1
+eclib:
+  - 20220621
 exiv2:
   - 0.27
 expat:


saraedum:
@isuruf: the recent eclib update blocked the automatic migration of sagelib for the Arb & FLINT upgrades. Adding eclib here should give us automatic migrations in the future for eclib (and seems to be the right thing anyway since it's a C++ library.)

hmaarrfk:
Can you turn this into a small migrator? 

https://github.com/saraedum/conda-forge-pinning-feedstock/blob/eclib/recipe/migrations/example.exyaml

copy this file into `eclib20220621.yaml` and fill it in accordingly. Also remove the pin from global config, we will have to catch the migrator as it finishes.

saraedum:
@hmaarrfk I could do that if it's really needed. However, currently, the only recipe that has eclib in its `host` section is `sagelib` (at least looking at the linux-64 repodata.) So maybe it's easier to just do that one manually?

PR #3430, opened 2022-09-15
pin google_cloud_cpp to the patch version

coryan:
I had forgotten that we change the inline namespace in each patch release.

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -476,7 +476,7 @@ glpk:
 gmp:
   - 6
 google_cloud_cpp:
-  - '2.2'
+  - '2.2.0'
 google_cloud_cpp_common:
   - 0.25.0
 googleapis_cpp:


hmaarrfk:
Your fix really needs to go in: https://github.com/conda-forge/google-cloud-cpp-feedstock/blob/main/recipe/meta.yaml

hmaarrfk:
Then, you might have to send repodata patches. Do you have many patch releases? Have people complained :/

coryan:
> Your fix really needs to go in: https://github.com/conda-forge/google-cloud-cpp-feedstock/blob/main/recipe/meta.yaml

Thanks. Will do.

> Then, you might have to send repodata patches.

Not sure what those are, but will be happy to send them when needed.

> Do you have many patch releases? Have people complained :/

No and no.  I am about to create a patch release, but we can ignore that for Conda: it basically changes some Bazel configuration, does not affect the code or packaging.

Sounds like I should close this PR then?

hmaarrfk:
Are you able to make this a new minor release?  Otherwise you may get unexpected errors. Moving forward, it should be fine and you won't need any patching.

coryan:
> Are you able to make this a new minor release?

No.

> Otherwise you may get unexpected errors. Moving forward, it should be fine and you won't need any patching.

I was planning to just ignore the patch release for Conda, is that possible?

hmaarrfk:
I think the best thing to do is:
1. Update the feedstock
2. Issue a repo data patch
3. Make a new patch release
4. Let the bot issue a new migrator

PR #3436, opened 2022-09-17
update patch versions for arrow-cpp

h-vetinari:
As part of the recent migrations, I updated the patch versions for [7.0.x](https://github.com/conda-forge/arrow-cpp-feedstock/commit/79abe2c85881262b58044c6a2bc75a94f7e21296) & [6.0.x](https://github.com/conda-forge/arrow-cpp-feedstock/commit/53ecc388f1fab8717341f8c8a942c8fff4bda378); follow-up to #3203

CC @conda-forge/arrow-cpp


h-vetinari:
AFAIU this should not need a migration...? If it does, I can create it.

kkraus14 (approved):

xhochy:
arrow-cpp has no ABI stability, so this should be a migration.

h-vetinari:
> arrow-cpp has no ABI stability, so this should be a migration.

Fair about the ABI, though these new versions should just be picked up with the next rerender of dependent feedstocks, and since we're building for these major versions already, the risk of breakage should be very very low...?

xhochy:
Yes

xhochy:
Still, it can lead to solving issues as run_exports is also down to the patch level because of the ABI instability.

h-vetinari:
> Still, it can lead to solving issues as run_exports is also down to the patch level because of the ABI instability.

Which is why I wanted to bump the patch version here, as a bunch of the recent migrations are only available for the new patch versions.

PR #3573, opened 2022-10-25
Add Python 3.11 migration

jakirkham:
Checklist

### DIFF
--- a/recipe/migrations/python311.yaml
+++ b/recipe/migrations/python311.yaml
@@ -1,0 +1,33 @@
+migrator_ts: 1666686085
+__migrator:
+    migration_number: 1
+    operation: key_add
+    primary_key: python
+    ordering:
+        python:
+            - 3.6.* *_cpython
+            - 3.7.* *_cpython
+            - 3.8.* *_cpython
+            - 3.9.* *_cpython
+            - 3.10.* *_cpython
+            - 3.11.* *_cpython  # new entry
+            - 3.6.* *_73_pypy
+            - 3.7.* *_73_pypy
+            - 3.8.* *_73_pypy
+            - 3.9.* *_73_pypy
+    paused: false
+    longterm: True
+    pr_limit: 5
+    exclude:
+      # this shouldn't attempt to modify the python feedstocks
+      - python
+      - pypy3.6
+      - pypy-meta
+
+python:
+  - 3.11.* *_cpython
+# additional entries to add for zip_keys
+numpy:
+  - 1.21
+python_impl:
+  - cpython
--- a/recipe/migrations/python311.yaml
+++ b/recipe/migrations/python311.yaml
@@ -28,6 +28,6 @@ python:
   - 3.11.* *_cpython
 # additional entries to add for zip_keys
 numpy:
-  - 1.21
+  - 1.23
 python_impl:
   - cpython
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -556,7 +556,6 @@ ntl:
 # we build for the oldest version possible of numpy for forward compatibility
 numpy:
   # part of a zip_keys: python, python_impl, numpy
-  - 1.20   # [not (osx and arm64)]
   - 1.20
   - 1.20
   - 1.21
@@ -614,13 +613,11 @@ pybind11_abi:
   - 4
 python:
   # part of a zip_keys: python, python_impl, numpy
-  - 3.7.* *_cpython    # [not (osx and arm64)]
   - 3.8.* *_cpython
   - 3.9.* *_cpython
   - 3.10.* *_cpython
 python_impl:
   # part of a zip_keys: python, python_impl, numpy
-  - cpython   # [not (osx and arm64)]
   - cpython
   - cpython
   - cpython


recipe/migrations/python311.yaml
```diff
@@ -0,0 +1,33 @@
...
+      # this shouldn't attempt to modify the python feedstocks
+      - python
+      - pypy3.6
+      - pypy-meta
+
+python:
+  - 3.11.* *_cpython
+# additional entries to add for zip_keys
+numpy:
+  - 1.21
```
xhochy, line 31: This should probably latest numpy. Older releases don't support 3.11. We can try to rebuild older ones, but we have never seen a benefit in past migrations.
```suggestion
  - 1.23
```

recipe/migrations/python311.yaml
```diff
@@ -0,0 +1,33 @@
+migrator_ts: 1666686085
+__migrator:
+    migration_number: 1
+    operation: key_add
+    primary_key: python
+    ordering:
+        python:
+            - 3.6.* *_cpython
```
xhochy, line 8: Do we need to list 3.6 and 3.7 here?

jakirkham: Yeah was wondering the same thing. Saw this in our PyPy migrator

https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/28885e3f7b38981651764e3dd966b3a21f5db818/recipe/migrations/pypy38.yaml#L8-L9

Also we had this in the Python 3.10 migrator ( https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/2008 ) ( even after we dropped Python 3.6, https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/1955 )

So thinking this is needed for other reasons. Hopefully someone can clarify. Also happy to drop if I've misunderstood

xhochy: Can you link a PR where this migration has been applied locally? If that works, all is fine.

jakirkham: Conda-smithy keeps trying to delete the migrator altogether (as it not merged in conda-forge-pinning). Forgetting the option to tell conda-smithy to override this behavior

mbargull: It shouldn't hurt to have superfluous entries. Pretty sure we can remove 3.6.
To remove 3.7 we need to https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/2623#issuecomment-1226073265 :
> \[...\] drop 3.7 when 3.11 comes out and we start the migration in October. \[...\]

As per the comment above, this PR should remove 3.7 from the current pinnings already (because AFAIR we currently don't have code to do add+remove in a single migration).

chrisburr: I've made a PR in https://github.com/conda-forge/pytest-feedstock/pull/152

(The option is `--exclusive-config-file  ../conda-forge-pinning-feedstock/recipe/conda_build_config.yaml`)

mbargull: > As per the comment above, this PR should remove 3.7 from the current pinnings already (because AFAIR we currently don't have code to do add+remove in a single migration).

(NB: If there surfaced new/other (good/important) reasons to drop it after the migration, that's fine by me. I was just reciting the comment/core decision from a few months back.)

jakirkham: Ah ok was looking for a way to do drop Python 3.7 in the migration, but didn't see it. Went ahead and deleted from the pinning file directly 👍

Thanks Chris! 🙏 Yeah had tried that option, but it was still deleting it for me. Probably missing something. Maybe it matters what the path to the pinning file is? In any event, looks like the PR is doing the right thing, which is encouraging 😄

recipe/migrations/python311.yaml
chrisburr: Given we're already using multiple numpy versions we might as well follow upstream.

recipe/migrations/python311.yaml
chrisburr: It needs to be able to find the correct `migrations/` folder next to `conda_build_config.yaml` so perhaps that was your issue?

chrisburr (approved):
I guess this can be unmarked as draft?

chrisburr:
Actually we probably need to fix crossenv before merging this to avoid lots of broken PRs: https://github.com/conda-forge/crossenv-feedstock/pull/34

chrisburr (approved):
I've cross-compiled `numpy` with this now so I think this can be merged

jakirkham:
cc @conda-forge/core

jakirkham:
Thanks Chris! Also thanks everyone for the reviews 🙏

recipe/migrations/python311.yaml
jakirkham: Oh I see. Thanks for clarifying! 🙏

Yeah probably. Was trying to point it to `conda_build_config.yaml` installed in `base` as opposed to a feedstock, which would not be adjacent to `migrations/` 😅

peastman:
Are python 3.11 packages supposed to be getting built now?  I'm trying to build new packages in https://github.com/conda-forge/openmm-feedstock/pull/95, but it still isn't building for 3.11.

PR #3619, opened 2022-10-30
Add gst_plugins_base to gstreamer migration

hmaarrfk:
I found this bug while going through the qt main migration.

https://github.com/conda-forge/qt-main-feedstock/pull/58

Maybe somebody can double check this fix for the migration. I will likely work around it for qt
Checklist

### DIFF
--- a/recipe/migrations/gstreamer121.yaml
+++ b/recipe/migrations/gstreamer121.yaml
@@ -1,9 +1,9 @@
 __migrator:
   build_number: 1
   kind: version
-  migration_number: 2
+  migration_number: 1
 gstreamer:
 - '1.21'
 gst_plugins_base:
 - '1.21'
-migrator_ts: 1667172469
+migrator_ts: 1664900131.838455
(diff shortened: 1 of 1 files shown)


recipe/migrations/gstreamer121.yaml
```diff
@@ -1,7 +1,9 @@
...
   build_number: 1
   kind: version
-  migration_number: 1
+  migration_number: 2
 gstreamer:
 - '1.21'
-migrator_ts: 1664900131.838455
+gst_plugins_base:
+- '1.21'
+migrator_ts: 1667172469
```
xhochy, line 9: Is this the case for all PRs? Otherwise, we don't need to restart the migration and a rerender on the respective PRs should be enough.

hmaarrfk: The PRs that went through should be ok. I'll undo the changes to the timestamp and migrator number.

xhochy (approved):

PR #3678, opened 2022-11-07
close out libprotobuf migration

h-vetinari:
The last two remaining feedstocks are:
* [tomviz](https://github.com/conda-forge/tomviz-feedstock/pull/49), which is stuck because upstream only(!) supports python 3.7
* [go-cockroach](https://github.com/conda-forge/go-cockroach-feedstock), an abandoned feedstock which already missed the last two protobuf migrations

I don't consider waiting for either of those a good reason to forestall this further.

CC @conda-forge/libprotobuf @hmaarrfk

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -472,7 +472,7 @@ libpcap:
 libpng:
   - 1.6
 libprotobuf:
-  - '3.20'
+  - '3.21'
 librdkafka:
   - '1.9'
 librsvg:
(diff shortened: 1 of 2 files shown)


hmaarrfk:
Yeah. I'm ok with this if these are indeed the last two. Is this a draft for an other reason?

h-vetinari:
> Yeah. I'm ok with this if these are indeed the last two. Is this a draft for an other reason?

It's not anymore. GH had opened it like this... 🙃

PR #3690, opened 2022-11-09
Add pytango to osx-arm64 migration

beenje:
Checklist

The stable version (9.3) of `pytango` isn't available for osx.
We added support in 9.4.0rc1 which was published in the `rc` branch: https://github.com/conda-forge/pytango-feedstock/tree/rc

We'd like to make it available for osx-arm64. Is it possible to create a migration for the rc branch? The PR target branch can maybe be changed after the PR creation? Or do we need to wait for this version to be available in main?

### DIFF
--- a/recipe/migrations/osx_arm64.txt
+++ b/recipe/migrations/osx_arm64.txt
@@ -863,6 +863,7 @@ cpptango
 tango-admin
 tango-database
 tango-test
+pytango
 omniorb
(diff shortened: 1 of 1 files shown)


xhochy:
If you want to have it on the `rc` branch, you will need to do the migration either manually (preferred) or add that branch to the `abi_migration_branches`.

PR #3731, opened 2022-11-17
bump to GCC 11

h-vetinari:
Last [update](https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/2010) was around a year ago, which fits with the yearly GCC cycle (though GCC 12 is in conda-forge already, if we want to).

I intentionally didn't bump the GCC version used in the CUDA builds, because AFAICT from [here](https://gist.github.com/ax3l/9489132) (xref also [here](https://github.com/conda-forge/nvcc-feedstock/issues/51)), CUDA 11.2 supports only up to GCC 10.


chrisburr (approved):
I'm not opposed to jumping to GCC 12 instead as I haven't heard of any notable issues with it.

hmaarrfk (approved):

hmaarrfk:
I might also want to jump to 12

kkraus14 (approved):
Another +1 to GCC 12

PR #3606, opened 2022-10-28
add migration for grpc_cpp 1.49

h-vetinari:
Similarly to the abseil migration https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/3181 recently, grpc-cpp has been renamed to libgrpc, see [here](https://github.com/conda-forge/grpc-cpp-feedstock/pull/245#issuecomment-1286759568).

I'm planning to raise PRs to all feedstocks that need upating.

CC @conda-forge/grpc-cpp 

Closes #3130

PS. This is going directly from 1.47 to 1.49 (1.50 has been published already, but we want to use that to work on making several deps [private](https://github.com/conda-forge/grpc-cpp-feedstock/pull/243), and it'll be good to have a separate version number to do the next migration once that's in place).

PR #3759, opened 2022-11-22
add missing libgrpc pin

h-vetinari:
Follow-up to #3606 / #3713; the former added an alias (now primary output), while the latter PR did not set that in the global cbc.

CC @xhochy @ocefpaf since you merged those two PRs


h-vetinari:
Fortunately this will not have any fall-out, because we haven't published anything for libgrpc 1.50 yet. :)

PR #3776, opened 2022-11-30
Add pyqt, pyqtwebengine, pyqtcharts to qt 5.15 migration

hmaarrfk:
It seems that some packages are pinned to 5.12 since they depend on pyqt and it isn't being migrated even though it has a run export of 'x.x'.

xref: https://github.com/conda-forge/lxml-feedstock/issues/75#issuecomment-1330774288

I think this should help those people move forward in the qt 5.15 migration.

Checklist

### DIFF
--- a/recipe/migrations/qt515.yaml
+++ b/recipe/migrations/qt515.yaml
@@ -8,3 +8,10 @@ qt:
   - 5.15
 qt_main:
   - 5.15
+# Upon closing, please rerember to add the pyqt to the global pinnings
+pyqt:
+  - 5.15
+pyqtwebengine:
+  - 5.15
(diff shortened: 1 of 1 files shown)


hmaarrfk:
I would appreciate somebody giving it an other set of eyes since I still don't fully understand "editing" migrations.

h-vetinari:
> I would appreciate somebody giving it an other set of eyes since I still don't fully understand "editing" migrations.

AFAIU (limited): as long as previous PRs in the migration don't need to be redone, it should be fine just to add it and the migrator will apply the newer version henceforth. Otherwise you need to bump the migration number.

PR #3785, opened 2022-12-01
Add migration for postgresql 15

hmaarrfk:
cc: @conda-forge/postgresql

This doesn't seem to have a current pinning, so many of my packages are defaulting to 14. 

- [ ] Should we "set the current pinning to 14"?

Checklist

### DIFF
--- a/recipe/migrations/libpq15.yaml
+++ b/recipe/migrations/libpq15.yaml
@@ -1,0 +1,12 @@
+migrator_ts: 1669902126
+__migrator:
+  kind: version
+  migration_number: 1
+  bump_number: 1
+
+libpq:
+  - 15
+postgresql:
+  - 15
+postgresql-plpython:
+  - 15
--- a/recipe/migrations/libpq15.yaml
+++ b/recipe/migrations/libpq15.yaml
@@ -8,5 +8,5 @@ libpq:
   - 15
 postgresql:
   - 15
-postgresql-plpython:
+postgresql_plpython:
   - 15


akrherz:
xref #133

hmaarrfk:
Thank you for that reference. I haven't seen sdvillal on conda-forge for some time. His input is missed :(. 

The truth is, that with the present usecase, we are effectively pinning people at some undefined version making packages more and more incompatible.

If we want to build out a matrix, I'm happy to do so. Maybe 14 / 15? It just starts to become a lot:

With the openssl migration, some packages would be built:
- 2 times for openssl
- 2 times for postgresql
- 4 times for python (3.8, 3.9, 3.10, 3.11)
- 3-6 times for different platforms.

hmaarrfk:
As a usecase example, qt5 just pinned itself to postgresql 15 https://github.com/conda-forge/qt-main-feedstock/pull/72

h-vetinari:
> If we want to build out a matrix, I'm happy to do so. Maybe 14 / 15? It just starts to become a lot:

I doubt that many projects would be affected by having both openssl and psql as a hostdep (and thus get the "x4" multiplication, rather than just x2 they already have from OpenSSL).

That said, do we have an idea how difficult it is for new psql versions to percolate through the ecosystem? If it's not that hard, we could just do regular migrators. But I guess to find out a double version migrator might be a good idea.

hmaarrfk:
Given that qt has already accidentally "migrated" to 15, I feel like the boat has already sailed on the challenges of moving to a new version of postgresql. 

Lets do a regular migration, and if there are challenges, we can do dual builds upon specific request.

PR #3814, opened 2022-12-07
add sdl_* and mpg123 to global pinning

### LINKED ISSUE #3809: Add SDL2 stack to pinning?
SDL2 is already in conda-forge; it'll gain importance because it's needed for [pygame](https://github.com/conda-forge/staged-recipes/pull/19665), which itself is needed for gym, ray, modin, etc.

So far, `sdl2_{image,mixer,ttf}` pinned `sdl2 2.0.*`, but that feedstock recently made a [big](https://github.com/conda-forge/sdl2-feedstock/pull/32) version jump; turns out that this is commensurate with what upstream [did](https://github.com/libsdl-org/SDL/tags?after=release-2.24.1), i.e. going from 2.0.22 -> 2.23.* -- combined with the fact that the pinning was 2.0.* previously -- very much makes it seem like we'd actually only need to pin to major level.

Looking at the [contents](https://github.com/regro/libcfgraph/blob/master/artifacts/sdl2/conda-forge/linux-64/sdl2-2.26.1-hf1af685_1.json#L215-L305) of sdl2, it's also pretty clear IMO that this falls within the purview of the global pinning.

`sdl2_{image,mixer}` are relatively current, but ttf is [woefully](https://github.com/libsdl-org/SDL_ttf) out of date. I plan to tackle this after https://github.com/conda-forge/sdl2_ttf-feedstock/pull/8.

That way, we could get rid of the ugly pins in the `sdl2_*` packages, and eventually migrate properly (SDL3 recently kicked off development).

Thoughts @conda-forge/sdl2 @conda-forge/sdl2_image @conda-forge/core?

h-vetinari:
Closes #3809

I've added run-exports to all the SDL2_* packages (and updated them where necessary; only pending https://github.com/conda-forge/sdl2_net-feedstock/pull/8), and verified [usages](https://github.com/search?q=org%3Aconda-forge%20-%20mpg123&type=code) of `mpg123` in conda-forge are all unpinned (and hence would pull in the newest one anyway).

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -533,6 +533,8 @@ mkl:
   - 2022
 mkl_devel:
   - 2022
+mpg123:
+  - '1.31'
 mpich:
   - 4
 mpfr:
@@ -658,6 +660,14 @@ ptscotch:
 s2n:
   - 1.3.28
 sdl2:
+  - '2'
+sdl2_image:
+  - '2'
+sdl2_mixer:
+  - '2'
+sdl2_net:
+  - '2'
+sdl2_ttf:
   - '2'
 singular:
   - 4.2.1.p3


recipe/conda_build_config.yaml
```diff
@@ -533,6 +533,8 @@ mkl:
   - 2022
 mkl_devel:
   - 2022
+mpg123:
+  - '1.31'
```
hmaarrfk, line 537: I only see https://github.com/conda-forge/yaafe-feedstock/pulls that needs mpg123, and they haven't rebuilt in a while.

Ok as is.

recipe/conda_build_config.yaml
```diff
@@ -659,6 +661,14 @@ s2n:
...
 sdl2:
   - '2'
+sdl2_image:
+  - '2'
+sdl2_mixer:
+  - '2'
+sdl2_net:
+  - '2'
+sdl2_ttf:
+  - '2'
```
hmaarrfk, line 671: hmmm. ok. i feel like we will forget to keep these in sync. but ok as well.

hmaarrfk (approved):

recipe/conda_build_config.yaml
h-vetinari: They shouldn't _necessarily_ be kept in sync. It just "happens" that upstream uses the same major number and version policy for all of them. Also, it doesn't get much clearer than having all libs named with the same prefix and grouped in the CBC 😂

PR #3824, opened 2022-12-09
Enable osx_arm64 for qt-3d|charts|datavis3d

jschueller:
cc @hmaarrfk


hmaarrfk:
Lets remember to update linux_aarch as well during the same migrator?

jschueller:
ok

PR #4003, opened 2023-01-27
remove wait_for_migrators from aws_c_*

h-vetinari:
Now that `aws_c_event_stream0218` is fully done, unblock next migrator from #3991, while we investigate https://github.com/regro/cf-scripts/issues/1595


h-vetinari:
Ugh, it seems like even after closing the migration completely (#4004), the migrator is still stuck on this.

h-vetinari:
Removing all the wait_for_migrator stuff due to https://github.com/regro/cf-scripts/issues/1595

PR #4059, opened 2023-02-05
Add gst_plugin_base pinning update to gstreamer plugin update

hmaarrfk:
I've already updated a few repos to no longer depend on the gst_plugin_base dependency:

* [x] https://github.com/conda-forge/gstlal-feedstock/pull/32
* [x] https://github.com/conda-forge/kivy-feedstock/pull/49
* [x] https://github.com/conda-forge/wxpython-feedstock/pull/97
* [x] https://github.com/conda-forge/wxwidgets-feedstock/pull/29
* [x] https://github.com/conda-forge/gst-libav-feedstock/pull/28

I think that maybe we can get rid of this extra pinning altogether to avoid these kinds of mistakes.

Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #4069, opened 2023-02-07
Drop openturns pinning

jschueller:
Rationale: this almost only affects feedstocks I maintain (ot* submodules depends on openturns) and end up getting in the way of my manual updates trigerring unnecessary MRs
For other feedstocks (batman, libcxx) their are not updated frequently at all, and our api has become more stable anyways so this has become unnecessary (and we use exports x.x version)

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -610,8 +610,6 @@ openmpi:
   - 4
 openssl:
   - '3'
-openturns:
-  - '1.20'
 orc:
   - 1.8.2
 pango:


hmaarrfk:
I guess you already have a workaround in 

https://github.com/conda-forge/otfftw-feedstock/blob/main/recipe/meta.yaml#L28

I've had to do something similar in https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/4059 to get things to move forward, but I think that the pinning, in general, is a good idea to keep, unless you want to loosen it to `x`.

jschueller:
I agree it's great in many cases, but not for this one where basically all the stack is updated at once

hmaarrfk:
I think that to ensure better collaboration between packages, if maintainers need to "opt out" of global pinnings (by setting it themselves locally as you've done) rather than "opt-in".

It seems that you have to likely add a manually pin to about 10 or so feedstocks, which to me is easier than removing the pinning for the one unsuspecting maintainer.

If your ABI (ABI stability here is more important) has become more stable, I think it would be more beneficial to update the pinning to `x`. On the next version, users will get `>=1.21,<2.a` (or something similar) requiring fewer rebuilds, and we can loosen the global pinning to `1` instead of  `1.x`

(Can we close #3515  ??)

jschueller:
yes please close #3515

jschueller:
you're right but its still annoying, maybe it could have been re-enabled later if openturns gets more rdepends ?

hmaarrfk:
sure. I do think that it would be good to loosen the pinning if you think that is appropriate.

PR #4215, opened 2023-03-11
GCC 12 & LLVM 15

h-vetinari:
Now that LLVM 15 has been [finally](https://github.com/conda-forge/clang-compiler-activation-feedstock/pull/91) made it to conda-forge (bit and pieces have been floating around for half a year, but we were missing some things to enable it), we could similarly update there as well (c.f. #2010 / #3290).

Already for https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/3731, a lot of people wanted to jump straight to GCC 12, so this should not be hugely controversial. If people want to wait some more for LLVM 15, I can split off the GCC changes into a separate PR.

Sidenote: LLVM 16 release is imminent (this time, the builds in conda-forge are [basically](https://github.com/conda-forge/clang-compiler-activation-feedstock/pull/95) ready 🥳), and GCC 13 is expected in April/May.

PS. Like for #3731, we can't use newer GCC together with CUDA 11.x compilers.


h-vetinari:
Since opening this PR, I found out there's _something_ happening in https://github.com/conda-forge/llvmdev-feedstock/pull/201 that causes the global `build.sh` and an output-specific install-script (having slightly different environments) to cause llvm to recompile on osx (rather than just install; possibly because something is being picked up as different between libcxx 14 & 15). We may want to figure that one out before we merge LLVM 15 as the default compiler. Any help/thoughts welcome.

chrisburr (approved):

h-vetinari:
> Since opening this PR, I found out there's _something_ happening in [conda-forge/llvmdev-feedstock#201](https://github.com/conda-forge/llvmdev-feedstock/pull/201) [...]

This should be fixed now (at least enough to not block the compiler bump, IMO) - thanks @isuruf!

h-vetinari:
With https://github.com/conda-forge/clang-compiler-activation-feedstock/pull/99 merged, we'd also have clang 16.0.0 available now, but I'd propose to let the dust settle on this a bit longer (e.g. 2-3 patch releases) before we make the switch. 15 should be good to go though.

h-vetinari:
> Are there current know issues with the transition?

None that I'm aware of. It's a bog-standard compiler version bump, nothing of substance should change, except of course the usual feature work, performance improvement, bug fixes etc.

> We seem to be hitting some challenges with libcxx [conda-forge/qt-main-feedstock#136 (comment)](https://github.com/conda-forge/qt-main-feedstock/pull/136#issuecomment-1489297116)

This is the fallout of the removal of a long-deprecated facility and therefore nothing unusual. I've commented there for available workarounds, but it's not an indication of problems in the stack.

h-vetinari:
So far this has two acks from core and no opposition so far (also not counting the additional 2 voices in support of moving to GCC 12 in #3731). Anything else that needs doing/discussing here @conda-forge/core?

hmaarrfk (approved):

h-vetinari:
Something I wasn't aware of for the to-do list before the next GCC bump: ensure that https://github.com/conda-forge/gfortran_osx-64-feedstock & https://github.com/conda-forge/gfortran_impl_osx-64-feedstock are up to date

jakirkham:
Should we add this to the conda-forge docs somewhere?

beckermr:
ah found it! here: https://conda-forge.org/docs/maintainer/infrastructure.html#compilers-and-runtimes

h-vetinari:
> ah found it! here: https://conda-forge.org/docs/maintainer/infrastructure.html#compilers-and-runtimes

Only that's highly outdated 😅

h-vetinari:
I can try to write something.

beckermr:
Great! It is good to keep historical information present in those docs though.

h-vetinari:
Done: https://github.com/conda-forge/conda-forge.github.io/pull/1946



> Great! It is good to keep historical information present in those docs though.

Well have to hash that out in the PR. I think text that describes a world based on the constraints of `comp7` (much less `comp4`) is noise that's detrimental to these docs being informative rather than confusing.

jakirkham:
Yeah this is a good point. Some maintainers want to be able to chip in, but need terse relevant info

The historical info is useful from a diagnosing odd behavior (or understanding the decisions made) perspective and maybe should be a separate section

How do other projects approach this separation of concerns?

PR #2847, opened 2022-05-05
Add `ucx` to global pinnings

jakirkham:
Start with the latest version 1.12.1, which added `run_exports`. Also relevant [ABI report]( https://abi-laboratory.pro/index.php?view=timeline&l=ucx ).

cc @conda-forge/ucx-split

xref: https://github.com/conda-forge/ucx-split-feedstock/pull/92/files#diff-f3725a55bf339595bf865fec73bda8ac99f283b0810c205442021f29c06eea9aR40-R41

<hr>

Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #4256, opened 2023-03-18
add ucx migration

h-vetinari:
ucx was added in https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/2847 and hasn't been updated since. Not sure what the bot is stumbling over, but it missed 1.13, and doesn't seem to picking up 1.14 either.

Following https://github.com/conda-forge/ucx-split-feedstock/pull/111, we should have a new migration (it's needed to re-enable ucx support e.g. in arrow).

CC @conda-forge/ucx-split @kkraus14 @pentschev

### DIFF
--- a/recipe/migrations/ucx1410.yaml
+++ b/recipe/migrations/ucx1410.yaml
@@ -1,0 +1,7 @@
+__migrator:
+  build_number: 1
+  kind: version
+  migration_number: 1
+ucx:
+- '1.14.0'
+migrator_ts: 1679152574.207815


jakirkham:
Thank you both! 🙏

jakirkham:
> ucx was added in https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/2847 and hasn't been updated since. Not sure what the bot is stumbling over, but it missed 1.13, and doesn't seem to picking up 1.14 either.

Raised issue ( https://github.com/regro/cf-scripts/issues/1880 )

PR #4464, opened 2023-05-17
fix google_cloud migrator

h-vetinari:
Follow-up to #4416


recipe/migrations/libgoogle_cloud2101.yaml
```diff
@@ -4,4 +4,6 @@ __migrator:
   migration_number: 1
 libgoogle_cloud:
 - 2.10.1
+google_cloud_cpp:
```
xhochy, line 7: I thought we weren't using this anymore?

h-vetinari:
> Due to #4465, you will need a new migrator.

No, it's OK, I picked that change into all the arrow PRs.

PR #4482, opened 2023-05-24
migrate libprotobuf 4.23 & grpc 1.55 (additionally to existing pins)

### LINKED ISSUE #4068: Dual `libprotobuf` pin
### Comment:

I have several environments that don't resolve with newer versions anymore because our `libprotobuf` pin is only supporting the `protobuf>=4` Python package. This causes several headaches for me as [`tritonclint-grpc`](https://github.com/conda-forge/tritonclient-split-feedstock/blob/55005affd42725347efdc4eedd2ea3130a14ef5e/recipe/meta.yaml#L64) (can be changed with the next release to `<4`) or [`streamlit`](https://github.com/conda-forge/streamlit-feedstock/blob/bdfab292988abc0f9221833410f1a7d76546c3f6/recipe/meta.yaml#L35) have a pin for `<4`. These are only two possible examples and it seems that many other packages are still on `protobuf=3`. This is not a problem in the Wheel world as there everyone ships a static protobuf build. 

Thus I would suggest to have two `libprotobuf` versions in our global pinning until the end of July 2023 (we need a point where we can drop it again, and half year's time from now seems reasonable).

cc @conda-forge/libprotobuf @conda-forge/core

### LINKED ISSUE #4075: Next round of abseil / grpc / protobuf upgrades
It's that time of the year again, it seems. We have:
* [x] https://github.com/conda-forge/abseil-cpp-feedstock/pull/53
* [x] https://github.com/conda-forge/grpc-cpp-feedstock/pull/266 / https://github.com/conda-forge/grpc-cpp-feedstock/pull/285
* [x] https://github.com/conda-forge/libprotobuf-feedstock/pull/144 / https://github.com/conda-forge/libprotobuf-feedstock/pull/156

in the pipeline. The first two should be straight-forward, though protobuf this time breaks the C++ API (a little), and adds a dependency on (the newest) abseil. Since `wait_for_migrators:` isn't [working](https://github.com/regro/cf-scripts/issues/1595) currently, I'd like to explicitly order the migrations as above, i.e. first abseil, then grpc, then protobuf.

This is also somewhat related to https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/4068, though I don't (yet) see that we'd need to build libprotobuf twice.

CC @conda-forge/abseil-cpp @conda-forge/grpc-cpp @conda-forge/libprotobuf 

PS. The first two are also dropping the compatibility outputs for the respective old names (`abseil-cpp` / `grpc-cpp`); hence we need to exclude them from the migrator, but I'd suggest to keep them in the pinning with the last available version, at least for a while longer.

h-vetinari:
Main discussion for this happened in #4075.

The situation:
* protobuf has a new version scheme where, the middle number is the feature level, and the "major" version corresponds to the language bindings -- e.g. v21.12 == v3.21.21 (C++) == v4.21.12 (python)
* the short-lived 22-series bumped the major version for C++ and also introduced a dependence on abseil, not just internally, but in the API. This will become [painful](https://github.com/conda-forge/abseil-cpp-feedstock/issues/45) later, but not now at least
* not only does it depend on abseil, but also abseil's test targets, which didn't compile yet with shared builds on windows and thus took a [while](https://github.com/conda-forge/abseil-cpp-feedstock/pull/58)
* libgrpc 1.55 is [apparently](https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/4075#issuecomment-1546655409) _only_ compatible with protobuf 4.23, hence the zip-keys here

Due to the potential fall-out from the major version bump, I'm suggesting to do a dual protobuf pin (see also https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/4068). This is fortuitous timing, because we've just completed a long-standing grpc-[refactor](https://github.com/conda-forge/grpc-cpp-feedstock/pull/243) as well (adding shared libs on windows, removing some run-exports for private libs), which will ease digesting that.

Closes #4068 
Closes #4075 
Closes #4436 
Closes #4454

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -172,6 +172,10 @@ zip_keys:
   -
     - arrow_cpp
     - libarrow
+  # until libprotobuf 3.21->4.23 migration is complete
+  -
+    - libgrpc
+    - libprotobuf
 
 
 # aarch64 specifics because conda-build sets many things to centos 6
--- a/recipe/migrations/protobuf423.yaml
+++ b/recipe/migrations/protobuf423.yaml
@@ -1,0 +1,11 @@
+__migrator:
+  build_number: 1
+  kind: version
+  migration_number: 1
+libgrpc:
+- '1.54'
+- '1.55'
+libprotobuf:
+- '3.21'
+- '4.23'
+migrator_ts: 1684932016.2362208


kkraus14 (approved):

kkraus14:
This looks reasonable to me given the whole situation with things.

h-vetinari:
> @h-vetinari are you tracking issues related to this?

I think it would be best to reopen #4075, but I don't have the rights to do that. Bazel needs https://github.com/conda-forge/grpc_java_plugin-feedstock/pull/57

PR #4527, opened 2023-06-01
Add message to maintainers about CUDA 12 transition

carterbox:
The purpose of this PR is to add more context to automigrations for CUDA 12 because the transition is more involved than just swapping out the compiler version.

Ideally, these notes would be added to the PR header, but I didn't see a key for that in the migration YAML. Maybe there is one that I don't know about?

Checklist

### DIFF
--- a/recipe/migrations/cuda120.yaml
+++ b/recipe/migrations/cuda120.yaml
@@ -45,14 +45,14 @@ __migrator:
       - 11.1                       # [(linux64 or win) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
       - 11.2                       # [(linux64 or win) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
       - 12.0                       # [(linux64 or win) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
-  commit_message: | Rebuild for CUDA 12
-
-The transition to CUDA 12 SDK includes new packages for all CUDA libraries and
-build tools. Notably, the cudatoolkit package no longer exists, and packages
-should depend directly on the specific cuda libraries (libblas, libcusolver,
-etc) as needed. A list of all new cuda packages and their status is available
-[in this issue](https://github.com/conda-forge/staged-recipes/issues/21382).
-
+  commit_message: |
+    Rebuild for CUDA 12
+    
+    The transition to CUDA 12 SDK includes new packages for all CUDA libraries and
+    build tools. Notably, the cudatoolkit package no longer exists, and packages
+    should depend directly on the specific cuda libraries (libblas, libcusolver,
+    etc) as needed. A list of all new cuda packages and their status is available
+    [in this issue]( https://github.com/conda-forge/staged-recipes/issues/21382 ).
 
 cuda_compiler:                 # [linux64 and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
   - cuda-nvcc                  # [linux64 and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
--- a/recipe/migrations/cuda120.yaml
+++ b/recipe/migrations/cuda120.yaml
@@ -5,7 +5,7 @@ __migrator:
   migration_number:
     1
   build_number:
-    1
+    2
   paused: false
   override_cbc_keys:
     - cuda_compiler_stub
--- a/recipe/migrations/cuda120.yaml
+++ b/recipe/migrations/cuda120.yaml
@@ -5,7 +5,7 @@ __migrator:
   migration_number:
     1
   build_number:
-    2
+    1
   paused: false
   override_cbc_keys:
     - cuda_compiler_stub
@@ -53,6 +53,7 @@ __migrator:
     should depend directly on the specific CUDA libraries (libblas, libcusolver,
     etc) as needed. For an in-depth overview of the changes and to report problems
     [see this issue]( https://github.com/conda-forge/conda-forge.github.io/issues/1963 ).
+    Please feel free to raise any issues encountered there. Thank you! :pray:
 
 cuda_compiler:                 # [linux64 and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
   - cuda-nvcc                  # [linux64 and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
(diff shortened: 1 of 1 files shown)


recipe/migrations/cuda120.yaml
```diff
@@ -45,7 +45,14 @@ __migrator:
...
       - 12.0                       # [(linux64 or win) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
-  commit_message: "Rebuild for CUDA 12"
+  commit_message: | Rebuild for CUDA 12
+
+The transition to CUDA 12 SDK includes new packages for all CUDA libraries and
+build tools. Notably, the cudatoolkit package no longer exists, and packages
+should depend directly on the specific cuda libraries (libblas, libcusolver,
+etc) as needed. A list of all new cuda packages and their status is available
+[in this issue](https://github.com/conda-forge/staged-recipes/issues/21382).
+
```
jakirkham, line 55:
```suggestion
  commit_message: |
    Rebuild for CUDA 12
    
    The transition to CUDA 12 SDK includes new packages for all CUDA libraries and
    build tools. Notably, the cudatoolkit package no longer exists, and packages
    should depend directly on the specific cuda libraries (libblas, libcusolver,
    etc) as needed. A list of all new cuda packages and their status is available
    [in this issue]( https://github.com/conda-forge/staged-recipes/issues/21382 ).
```

jakirkham (commented):
Thanks Daniel! 🙏 

Think we need to tweak the formatting for YAML to handle this correctly

Also changed `libblas` to `libcublas` and spaced out the link (to avoid rendering issues)

recipe/migrations/cuda120.yaml
jakirkham: FWIW tested this with [this online JSON to YAML converting tool]( https://jsonformatter.org/yaml-to-json ) to confirm it grouped this in one string under `commit_message`

carterbox:
I feel like there may be other useful information to add to this text.

For example, adding something about the headers moving around? It seems that the headers files for `libcublas-dev` are located in `$PREFIX/targets/x86-linux/include` instead of `$PREFIX/include`. I'm not sure whether this intentionally breaks packaging convention for our channel.

jakirkham:
With previous CUDA versions, the headers came from the Docker image (so were not in `$PREFIX/include`)

In CUDA 12, the headers are supplied. However they are supplied in architecture specific directories to enable cross-compilation (amongst other things)

Some context in issue ( https://github.com/conda-forge/cuda-nvcc-feedstock/issues/12 ). Though there were other discussions prior

jakirkham:
If you are ok with the suggestion above, perhaps we can commit that and then discuss changes on top of the new PR diff?

Would be good to address the YAML formatting issues first (since those unfortunately affect the diff of the whole message 😞)

carterbox:
I don't really know what message to write about how the headers are supplied because after reading through some Issues/PRs, it's still not clear to me whether downstream maintainers are expected to add the new include path or if changes to CUDA packages are forthcoming.

carterbox:
Sounds good to me, thanks for following the migration so closely!

jakirkham:
Ok have filed issue ( https://github.com/conda-forge/conda-forge.github.io/issues/1963 )

Please look it over and let me know if you have questions. Happy to iterate on the content there and clarify anything else as needed 🙂

jakirkham:
Also thanks for diving in here and providing useful feedback! 🙏

carterbox:
I think the bringup issue is very helpful and is the best place to point maintainers.

recipe/migrations/cuda120.yaml
```diff
@@ -5,7 +5,7 @@ __migrator:
   migration_number:
     1
   build_number:
-    1
+    2
```
jakirkham, line 8: Think this is just for the build number increment in PRs. Would it be alright if we kept this?

```suggestion
    1
```

recipe/migrations/cuda120.yaml
```diff
@@ -45,7 +45,14 @@ __migrator:
...
       - 12.0                       # [(linux64 or win) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
-  commit_message: "Rebuild for CUDA 12"
+  commit_message: |
+    Rebuild for CUDA 12
+    
+    The transition to CUDA 12 SDK includes new packages for all CUDA libraries and
+    build tools. Notably, the cudatoolkit package no longer exists, and packages
+    should depend directly on the specific CUDA libraries (libblas, libcusolver,
+    etc) as needed. For an in-depth overview of the changes and to report problems
+    [see this issue]( https://github.com/conda-forge/conda-forge.github.io/issues/1963 ).
```
jakirkham, line 55:
```suggestion
    [see this issue]( https://github.com/conda-forge/conda-forge.github.io/issues/1963 ).
    Please feel free to raise any issues encountered there. Thank you! :pray:
```

jakirkham (commented):
Thanks Daniel! 🙏

This looks good. Had some minor tweaks.

Also thanks for reading through the CUDA 12 bringup issue ❤️

recipe/migrations/cuda120.yaml
carterbox: 🙏🏼

recipe/migrations/cuda120.yaml
carterbox: 🤷🏼 I am honestly not sure what this parameter does.

jakirkham:
Thanks Daniel! 🙏

jakirkham:
Looks like the whole commit message gets added to the PR title

Here's an example PR ( https://github.com/conda-forge/timemory-feedstock/pull/15 ). Also added a screenshot below

Screen Shot 2023-06-01 at 10 39 11 PM

Submitting PR ( https://github.com/regro/cf-scripts/pull/1699 ) upstream to fix this

jakirkham:
Think we can improve the visibility of commit messages (like this one) by adding it to the PR body. Have tried to do this with PR ( https://github.com/regro/cf-scripts/pull/1713 )

jakirkham:
Here's an example with the aforementioned update:

Screenshot 2023-06-14 at 7 32 15 PM

jakirkham:
Think we can tweak the formatting slightly to make the message stick out a bit more ( https://github.com/regro/cf-scripts/pull/1715)

jakirkham:
After those last changes, here's a refreshed example ( https://github.com/conda-forge/rmm-feedstock/pull/25 ):

Screenshot 2023-06-21 at 12 43 36 AM

jakirkham:
Have also edited the OPs of the still open PRs that preceded this change to include the updated messaging. Hopefully that provides more context for those maintainers when they are able to look

PR #4534, opened 2023-06-01
In no CUDA case, set `cuda*` variables `undefined`

jakirkham:
Instead of setting `cuda*` variables to `None` in the no CUDA (CPU case), set them to `undefined`. This works better with Jinja filters (`|`), which makes it easier to work with.

<hr>

Checklist

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -44,19 +44,19 @@ m2w64_fortran_compiler:        # [win]
   - m2w64-toolchain            # [win]
 
 cuda_compiler:
-  - None
+  - undefined
   - nvcc                       # [(linux64 or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
   - nvcc                       # [(linux64 or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
   - nvcc                       # [(linux64 or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
   - nvcc                       # [(linux64 or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
 cuda_compiler_version:
-  - None
+  - undefined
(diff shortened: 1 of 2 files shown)


mikemhenry:
I think that would make things like `.startswith` work cleaner, but people will be grumpy with this breaking old version selectors, but I am in favor of that since going forward it makes things easier.

jakirkham:
Well said!

IIUC think it would be using `|startswith` (Jinja handles `undefined` as `""` in that case unlike `.startswith`)

<hr>

Just to contextualize this, a line like this...

```yaml
- libcublas-dev      # [(cuda_compiler_version or "").startswith("12")]
```

...could become...


```yaml
- libcublas-dev      # [cuda_compiler_version|startswith("12")]
```

mikemhenry:
I forgot, one other foot gun is that if anyone is treating `None` as false, changing it to a string will make it truthy, so it could silently break things.

I am all for this change, just trying to enumerate some of the edge cases that will make people grumpy when their feedstocks get busted.

jakirkham:
Yeah think the main issue is when recipes compare to `None` explicitly. Seems there are [a few cases of this]( https://github.com/search?q=org%3Aconda-forge+path%3Arecipe+%22cuda_compiler_version%22+%22None%22&type=code )

mikemhenry:
For me as a feedstock maintainer, if the migration bot for this included an explanation of what changed, and a few examples on how to migrate, I think it would be helpful. 

But since this isn't a migration, I think people will re-render and not get any explanation on what changed and will just be confused.

jakirkham:
That's true we could put this in a migrator instead. Also could include a custom commit message with a detail about what changed

mikemhenry:
I *think* that would be better so there is less of a rug pull

jakirkham:
It's worth noting the migrator PR title is taken from [the first line of the commit message]( https://github.com/regro/cf-scripts/blob/927e6b295088911a3023d337e65ef3e082a45701/conda_forge_tick/migrators/migration_yaml.py#L336 ) so we can make this change pretty visible

PR #4605, opened 2023-06-27
Soft close ffmpeg 6 migration stopping the dual builds

hmaarrfk:
I've checked that the remaining feedstocks fore the FFMPEG migration are the same ones that were remaining for the ffmpeg 5 migration. They all seem to need upstream changes and many of the feedstocks seem to be abandoned.

@h-vetinari @conda-forge/opencv this should help avoid the huge build matrix in https://github.com/conda-forge/opencv-feedstock/pull/363#issuecomment-1608672751

Checklist

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -333,7 +333,7 @@ exiv2:
 expat:
   - 2
 ffmpeg:
-  - '5'
+  - '6'
 fftw:
   - 3
 flann:
--- a/recipe/migrations/ffmpeg6.yaml
+++ b/recipe/migrations/ffmpeg6.yaml
@@ -3,6 +3,5 @@ __migrator:
   kind: version
   migration_number: 1
 ffmpeg:
-- '5'
 - '6'
 migrator_ts: 1684252387.6420875


h-vetinari:
We've got less than 2/3rds of ffmpeg dependencies done, these are open as of now:
* https://github.com/conda-forge/mlt-feedstock/pull/22
* https://github.com/conda-forge/openscenegraph-feedstock/pull/28
* https://github.com/conda-forge/openimageio-feedstock/pull/77
* https://github.com/conda-forge/xpra-feedstock/pull/18
* https://github.com/conda-forge/unpaper-feedstock/pull/5
* https://github.com/conda-forge/lintel-feedstock/pull/23
* https://github.com/conda-forge/aubio-feedstock/pull/37
* https://github.com/conda-forge/gepetto-viewer-feedstock 

Plus two (kfilemetadata, kdenlive) from https://github.com/conda-forge/conda-forge.github.io/issues/1861 that I'm not counting.

I don't know if it's reasonable to close the migration already, but I don't have anything against it. There's another way to cut down the insane opencv build matrix, but it'll take a few days yet. If we can live with the big build matrix in the meantime, I don't see that we have to close yet, but again, that's alright as well.

hmaarrfk:
I guess we are at the point where I would disagree that the "migration" is still live.

All the aforementioned  packages have one or multiple other standing migrations.

In my mind, these "leaves" of the ecosystem should be allowed to be "shed".

If a new package requires them they are strongly incentivized to update the packages that depend on ffmpeg to modernize them.

If a new package doesn't require them, then why should they build against an old version of the ffmpeg library only to get migrated immediately.

The bot can help these slower packages get caught up, but there is no reason to hold back all the other non-leaves.

h-vetinari:
Fair enough, SGTM!

PR #4620, opened 2023-06-30
restart new migration for protobuf 4.23.3 & grpc 1.56

h-vetinari:
As [discussed](https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/4075#issuecomment-1614186341) in #4075

Closes #4618 
Closes #4617

### DIFF
--- a/recipe/migrations/libprotobuf4233_libgrpc156.yaml
+++ b/recipe/migrations/libprotobuf4233_libgrpc156.yaml
@@ -1,11 +1,11 @@
 __migrator:
   build_number: 1
   kind: version
-  migration_number: 2
+  migration_number: 1
 libgrpc:
 - '1.54'
-- '1.55'
+- '1.56'
 libprotobuf:
 - '3.21'
-- '4.23.2'
-migrator_ts: 1684932016.2362208
+- '4.23.3'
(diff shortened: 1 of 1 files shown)


recipe/migrations/libprotobuf4233_libgrpc156.yaml
```diff
@@ -1,11 +1,11 @@
 __migrator:
   build_number: 1
   kind: version
-  migration_number: 2
+  migration_number: 1
```
xhochy, line 4: Shouldn't we increase the number instead of decreasing it?

xhochy: Or is this different as it stops one and starts another migration?

h-vetinari: I renamed the migration, so the PR names are less confusing. Rather than a 3rd "Rebuild for protobuf423", it would be "Rebuild for libprotobuf4233_libgrpc156".

h-vetinari: > Or is this different as it stops one and starts another migration?

Exactly

PR #4744, opened 2023-08-06
Close out five Poppler migrations.

pkgw:
These have largely been stuck on `kfilemetadata`, which is super out-of-date and does not appear likely to get updated any time soon. I also just finished up https://github.com/conda-forge/gimagereader-qt-feedstock/pull/15 , which is actually maintainable but also needed some work (leptonica → tesseract → gimagereader-qt).

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -629,7 +629,7 @@ pixman:
 poco:
   - 1.12.4
 poppler:
-  - '22.12'
+  - '23.07'
 postgresql:
   - 15
 postgresql_plpython:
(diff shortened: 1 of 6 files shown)


h-vetinari:
> These have largely been stuck on `kfilemetadata`

I still think we should archive the k* feedstocks: https://github.com/conda-forge/conda-forge.github.io/issues/1861. No-one is touching those, and they just get in the way of migrations constantly.

pkgw:
@h-vetinari I don't have a sense as to how much anyone depends on those packages in the wild, but I basically agree.

PR #4834, opened 2023-08-23
Add CUDA 11.8 migrator

jakirkham:
Adds a migrator to include CUDA 11.8 as discussed in issue ( https://github.com/conda-forge/conda-forge.github.io/issues/1981 ).

<hr>

Checklist

Fixes https://github.com/conda-forge/conda-forge.github.io/issues/1981

Closes https://github.com/conda-forge/ucx-split-feedstock/pull/133
Closes https://github.com/conda-forge/nccl-feedstock/pull/96
Closes https://github.com/conda-forge/pycolmap-feedstock/pull/7

### DIFF
--- a/recipe/migrations/cuda118.yaml
+++ b/recipe/migrations/cuda118.yaml
@@ -89,7 +89,7 @@ fortran_compiler_version:      # [linux and os.environ.get("CF_CUDA_ENABLED", "F
 cdt_name:                      # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
   - cos7                       # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
 
-docker_image:                                      # [os.environ.get("BUILD_PLATFORM", "").startswith("linux-") and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+docker_image:                                         # [os.environ.get("BUILD_PLATFORM", "").startswith("linux-") and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
   - quay.io/condaforge/linux-anvil-cuda:11.8          # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
   # case: native compilation (build == target)
   - quay.io/condaforge/linux-anvil-ppc64le-cuda:11.8  # [ppc64le and os.environ.get("BUILD_PLATFORM") == "linux-ppc64le"]
--- a/recipe/migrations/cuda118.yaml
+++ b/recipe/migrations/cuda118.yaml
@@ -58,7 +58,7 @@ __migrator:
       - quay.io/condaforge/linux-anvil-cuda:11.8          # [ppc64le and os.environ.get("BUILD_PLATFORM") == "linux-64"]
       - quay.io/condaforge/linux-anvil-cuda:11.8          # [aarch64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
 
-      # non-CUDA arch: cross-compilation (build != target)
+      # Native CentOS 7 image
       - quay.io/condaforge/linux-anvil-cos7-x86_64        # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
     cuda_compiler_version:
       - None
--- a/recipe/migrations/cuda118.yaml
+++ b/recipe/migrations/cuda118.yaml
@@ -1,0 +1,91 @@
+migrator_ts: 1692828152
+__migrator:
+  kind:
+    version
+  migration_number:
+    1
+  build_number:
+    1
+  paused: false
+  override_cbc_keys:
+    - cuda_compiler_stub
+  operation: key_add
+  check_solvable: false
+  primary_key: cuda_compiler_version
+  ordering:
+    cxx_compiler_version:
+      - 9
+      - 8
+      - 7
+    c_compiler_version:
+      - 9
+      - 8
+      - 7
+    fortran_compiler_version:
+      - 9
+      - 8
+      - 7
+    docker_image:
+      - quay.io/condaforge/linux-anvil-comp7              # [os.environ.get("BUILD_PLATFORM") == "linux-64"]
+      - quay.io/condaforge/linux-anvil-aarch64            # [os.environ.get("BUILD_PLATFORM") == "linux-aarch64"]
+      - quay.io/condaforge/linux-anvil-ppc64le            # [os.environ.get("BUILD_PLATFORM") == "linux-ppc64le"]
+      - quay.io/condaforge/linux-anvil-armv7l             # [os.environ.get("BUILD_PLATFORM") == "linux-armv7l"]
+      - quay.io/condaforge/linux-anvil-cuda:9.2           # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
+      - quay.io/condaforge/linux-anvil-cuda:10.0          # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
+      - quay.io/condaforge/linux-anvil-cuda:10.1          # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
+      - quay.io/condaforge/linux-anvil-cuda:10.2          # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
+      - quay.io/condaforge/linux-anvil-cuda:11.0          # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
+      - quay.io/condaforge/linux-anvil-cuda:11.1          # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
+      - quay.io/condaforge/linux-anvil-cuda:11.2          # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
+      # case: native compilation (build == target)
+      - quay.io/condaforge/linux-anvil-ppc64le-cuda:11.2  # [ppc64le and os.environ.get("BUILD_PLATFORM") == "linux-ppc64le"]
+      - quay.io/condaforge/linux-anvil-aarch64-cuda:11.2  # [aarch64 and os.environ.get("BUILD_PLATFORM") == "linux-aarch64"]
+      # case: cross-compilation (build != target)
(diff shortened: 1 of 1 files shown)


jakirkham:
Started a couple test PRs using this migration:

* https://github.com/conda-forge/ucx-split-feedstock/pull/133
* https://github.com/conda-forge/pycolmap-feedstock/pull/7

jakirkham:
It appears that native builds work ok. However CUDA 11.8 cross-compilation support is absent atm. Started PR ( https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/261 ) to add CUDA 11.8 cross-compilation support

recipe/migrations/cuda118.yaml
```diff
@@ -0,0 +1,91 @@
...
+      - quay.io/condaforge/linux-anvil-cuda:11.0          # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
+      - quay.io/condaforge/linux-anvil-cuda:11.1          # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
+      - quay.io/condaforge/linux-anvil-cuda:11.2          # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
+      # case: native compilation (build == target)
+      - quay.io/condaforge/linux-anvil-ppc64le-cuda:11.2  # [ppc64le and os.environ.get("BUILD_PLATFORM") == "linux-ppc64le"]
+      - quay.io/condaforge/linux-anvil-aarch64-cuda:11.2  # [aarch64 and os.environ.get("BUILD_PLATFORM") == "linux-aarch64"]
+      # case: cross-compilation (build != target)
+      - quay.io/condaforge/linux-anvil-cuda:11.2          # [ppc64le and os.environ.get("BUILD_PLATFORM") == "linux-64"]
+      - quay.io/condaforge/linux-anvil-cuda:11.2          # [aarch64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
+      - quay.io/condaforge/linux-anvil-cuda:11.8          # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
```
h-vetinari, line 46: The last line does not look like cross-compilation, but I think it's just an issue of how the lines are C&P'd together. Probably good to either reorder the lines or add another comment like below
```suggestion
      # case: cross-compilation (build != target)
      - quay.io/condaforge/linux-anvil-cuda:11.2          # [ppc64le and os.environ.get("BUILD_PLATFORM") == "linux-64"]
      - quay.io/condaforge/linux-anvil-cuda:11.2          # [aarch64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
      # CUDA 11.8
      - quay.io/condaforge/linux-anvil-cuda:11.8          # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
```

h-vetinari (commented):
Basically LGTM!

recipe/migrations/cuda118.yaml
jakirkham: Think we are stuck with the order. That said, the point about adding some comments is well-taken

recipe/migrations/cuda118.yaml
```diff
@@ -0,0 +1,99 @@
...
+      # Native builds
+      - quay.io/condaforge/linux-anvil-comp7              # [os.environ.get("BUILD_PLATFORM") == "linux-64"]
+      - quay.io/condaforge/linux-anvil-aarch64            # [os.environ.get("BUILD_PLATFORM") == "linux-aarch64"]
+      - quay.io/condaforge/linux-anvil-ppc64le            # [os.environ.get("BUILD_PLATFORM") == "linux-ppc64le"]
+      - quay.io/condaforge/linux-anvil-armv7l             # [os.environ.get("BUILD_PLATFORM") == "linux-armv7l"]
+
+      # Legacy CUDAs
+      - quay.io/condaforge/linux-anvil-cuda:9.2           # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
+      - quay.io/condaforge/linux-anvil-cuda:10.0          # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
+      - quay.io/condaforge/linux-anvil-cuda:10.1          # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
```
leofang, line 38: Q: Do we still need these? I thought they are not zippable because `cuda_compiler_version` does not list them?

recipe/migrations/cuda118.yaml
jakirkham: Added some comments below. Please let me know what you think

recipe/migrations/cuda118.yaml
```diff
@@ -26,31 +26,39 @@ __migrator:
...
       - quay.io/condaforge/linux-anvil-ppc64le-cuda:11.8  # [ppc64le and os.environ.get("BUILD_PLATFORM") == "linux-ppc64le"]
       - quay.io/condaforge/linux-anvil-aarch64-cuda:11.8  # [aarch64 and os.environ.get("BUILD_PLATFORM") == "linux-aarch64"]
-      # case: cross-compilation (build != target)
+      # CUDA 11.8 arch: cross-compilation (build != target)
       - quay.io/condaforge/linux-anvil-cuda:11.8          # [ppc64le and os.environ.get("BUILD_PLATFORM") == "linux-64"]
       - quay.io/condaforge/linux-anvil-cuda:11.8          # [aarch64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
-      # case: non-CUDA builds
+
+      # non-CUDA arch: cross-compilation (build != target)
       - quay.io/condaforge/linux-anvil-cos7-x86_64        # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
```
h-vetinari: Comments look good in general, but this last one is definitely not cross-compilation. I also don't know why we even have non-CUDA at the end again, given that this should be covered by the lines at the beginning.

recipe/migrations/cuda118.yaml
```diff
@@ -0,0 +1,99 @@
...
+      - quay.io/condaforge/linux-anvil-cuda:11.8          # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
+      # CUDA 11.8 arch: native compilation (build == target)
+      - quay.io/condaforge/linux-anvil-ppc64le-cuda:11.8  # [ppc64le and os.environ.get("BUILD_PLATFORM") == "linux-ppc64le"]
+      - quay.io/condaforge/linux-anvil-aarch64-cuda:11.8  # [aarch64 and os.environ.get("BUILD_PLATFORM") == "linux-aarch64"]
+      # CUDA 11.8 arch: cross-compilation (build != target)
+      - quay.io/condaforge/linux-anvil-cuda:11.8          # [ppc64le and os.environ.get("BUILD_PLATFORM") == "linux-64"]
+      - quay.io/condaforge/linux-anvil-cuda:11.8          # [aarch64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
+
+      # non-CUDA arch: cross-compilation (build != target)
+      - quay.io/condaforge/linux-anvil-cos7-x86_64        # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
```
leofang, line 62: Q: Wouldn't this last item be zipped with `cuda_compiler_version` == `12.0`? (I guess I never understood the meaning of `ordering` in the migrator, even after the CUDA 12 migration. Would be nice if someone could shed some light 🙂)

leofang (changes requested):

recipe/migrations/cuda118.yaml
leofang: (My very same question from the old days: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/4400#discussion_r1188782828)

recipe/migrations/cuda118.yaml
leofang: It seems we're asking the same question (https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/4834#discussion_r1312659535) 🙂

recipe/migrations/cuda118.yaml
jakirkham: They didn't seem to cause any issues when applying the migration to different feedstocks

Zipping isn't relevant here as this is the `ordering` section and is only used to preserve the order of these when they do appear

recipe/migrations/cuda118.yaml
jakirkham: Applied this to feedstocks that completed the CUDA 12 migration (so they had both migrators like NCCL). AFAICT it worked without issues

This conda-build sorts some keys (so loses the ordering). As a result we have to reorder things after the fact in conda-smithy

In any event as the long term goal is to eliminate the need for different Docker images (as is already happening with CUDA 12.0). Think the `docker_image` ordering can eventually go away. So don't think it is worth thinking to hard about

recipe/migrations/cuda118.yaml
jakirkham: Fixed the comment. Sorry about that

It appears to be last in the CUDA 12.0 migrator where this was taken from

https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/33db02d50ade2f110a9420d5600dc8d258576a58/recipe/migrations/cuda120.yaml#L46-L47

h-vetinari: I know that this is taken from the CUDA 12 one, but I didn't get why it was necessary. In any case, not a blocker.

h-vetinari (approved):
LGTM thx!

leofang (approved):

recipe/migrations/cuda118.yaml
jakirkham: Tried to improve the comments in the global pinnings as well with PR: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/4865

jakirkham:
Thanks all! 🙏

Going to go ahead and get this started then. We can follow up on anything else separately

leofang:
I noticed that `cusparselt` is marked migrated, but it isn't, likely because it has this skip condition
```
skip: true  # [ppc64le or cuda_compiler_version not in ("11.2", "12.0")]
```
and the bot can't see any diff before/after applying the migrator. This would likely happen to other feedstocks like `cupy` too.

jakirkham:
Are there differences in the `cusparselt` binary needed for the CUDA 11.8 case?

PR #4895, opened 2023-09-11
Add openmp to arch_rebuild.txt

yangxingyu588:
Checklist

l need to build  `fastani` on Linux aarch64 but currently they fail with the following error:
```
fastani
              -nothing provides openmp
```
l hope that with the proposed modifications in this PR they will be usable on LInux aarch64

### DIFF
--- a/recipe/migrations/arch_rebuild.txt
+++ b/recipe/migrations/arch_rebuild.txt
@@ -378,6 +378,7 @@ openai
 openblas
 opencv
 openmm
+openmp
 openmpi
 openslide
 openslide-python


yangxingyu588:
@conda-forge-admin , please rerender

yangxingyu588:
> where is the `fastani` feedstock
that in [bioconda](https://github.com/bioconda/bioconda-recipes/tree/master/recipes/fastani), but [openmp ](https://github.com/conda-forge/openmpi-feedstock)in here

yangxingyu588:
is there anything else that needs to be modified?

hmaarrfk:
i'm just kinda confused as to why this was never built before. is this package even needed for those architectures?

hmaarrfk:
@conda-forge/openmp

h-vetinari:
> l need to build `fastani` on Linux aarch64 but currently they fail with the following error:
> 
> ```
> fastani
>               -nothing provides openmp
> ```
> 
> l hope that with the proposed modifications in this PR they will be usable on Linux aarch64

The output `openmp` has seen no new packages since [4+ years](https://anaconda.org/conda-forge/openmp/files). You'll want `llvm-openmp` (which is the new name), which already has linux-aarch builds for a long time.

As you might now, there are several OpenMP implementations; for example, on linux you could also use the `libgomp` output for GCC's (instead of LLVM's) implementation. However:
* building a library against one OpenMP flavour does not mean the resulting binary is compatible with another OpenMP implementation
* there can only be one OpenMP in an environment 

You may want to take this into account in how you devise your builds (i.e. building fastani against `llvm-openmp` means it cannot be installed side-by-side with something requiring `libgomp`), but since you're currently already relying on the LLVM-variant, staying with just that is probably alright.

h-vetinari:
PS. The PR wouldn't change anything, as the [openmp](https://github.com/conda-forge/openmp-feedstock) feedstock[^1] already builds for linux-aarch64.

[^1]: the list contains feedstock names, not output names

h-vetinari:
I've explained above why this PR doesn't do anything, and how to solve the issue - closing.

PR #4978, opened 2023-10-02
add max_solver_attempts to boost migration

h-vetinari:
I don't know what the default behaviour is, but there was an outage (causing package uploads to fail) early in the migration, and the bot hasn't retried any of those despite the packages being there now.


beckermr:
Default is 3. Bot tries prs that have never been tried before circling back to failed solver attempts.

h-vetinari:
> Default is 3. Bot tries prs that have never been tried before circling back to failed solver attempts.

Ah OK; that means if there's a long queue of "waiting for PR" feedstocks, it won't get retried yet...? That's unfortunate, because that means we'll go through all of the depth=0 and depth=1 feedstocks before we get back to the failed ones with much larger depth of children.

PR #4984, opened 2023-10-03
switch python 3.12 migration to use main channel

h-vetinari:
https://github.com/conda-forge/python-feedstock/pull/651 was merged, so we don't need the rc-channel in the migrator anymore.

### DIFF
--- a/recipe/migrations/python312.yaml
+++ b/recipe/migrations/python312.yaml
@@ -28,13 +28,9 @@ __migrator:
       - cross-python
       - python_abi
     exclude_pinned_pkgs: false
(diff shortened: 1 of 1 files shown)


ocefpaf (approved):

jakirkham:
Would the migration need to be restarted with this change or is that not needed for some reason?

h-vetinari:
Not needed. A migration only needs restarting if the builds before/after restart are incompatible with each other. In this case, the main point was that the ABI won't change between rc3 and GA, so we can build packages beforehand and know they'll be compatible afterwards.

PR #5113, opened 2023-11-01
next round of grpc/protobuf migration

h-vetinari:
Nothing left but the usual stragglers (already for a couple of weeks), let's move on before protobuf 4.25 arrives.

Closes #5024 
Closes #4998


hmaarrfk:
Thanks!

PR #5173, opened 2023-11-20
add pin for tensorflow

njzjz:
Checklist

ping @conda-forge/tensorflow 

TensorFlow has API compatibility but does not have ABI compatibility. This means the library linked to the TensorFlow C++ libraries needs to be rebuilt for every TensorFlow version. So in this PR, I added this pin to make things automatic.

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -762,6 +762,8 @@ tbb:
   - '2021'
 tbb_devel:
   - '2021'
+tensorflow:
+  - '2.14'
 thrift_cpp:
   - 0.18.1
 tinyxml2:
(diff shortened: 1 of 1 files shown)


xhochy (commented):
Shouldn't we also pin `libtensorflow` and `libtensorflow_cc`?

njzjz:
> Shouldn't we also pin `libtensorflow` and `libtensorflow_cc`?

Added. I am not sure whether they should be added to `zip_keys`.

h-vetinari:
> I am not sure whether they should be added to `zip_keys`.

Zip keys don't really work well for single entries, so it's fine not to add.

njzjz:
It seems @regro-cf-autotick-bot does not automatically update the pining...

xhochy:
@conda-forge/bot Any ideas where to look?

h-vetinari:
> Any ideas where to look?

I think the bot (in https://github.com/regro/cf-scripts) only looks at packages that have the same name as the feedstock, but I haven't delved into this. It's an issue for several multi-output recipes (especially where packages got renamed away from the original feedstockname).

PR #5336, opened 2024-01-03
tighten pin in fmt migration

h-vetinari:
Fmt 10.2 turned out not to be ABI compatible, see https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/628 and linked issues.

### DIFF
--- a/recipe/migrations/fmt10.yaml
+++ b/recipe/migrations/fmt10.yaml
@@ -3,7 +3,7 @@ __migrator:
   kind: version
   migration_number: 1
 fmt:
-- '10'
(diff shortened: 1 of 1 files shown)


bdice (approved):

h-vetinari:
Marking as draft because it might turn out that the ABI-breakage was a bug, rather than intentional.

h-vetinari:
This PR is now obsolete, because we'll be able to keep the major-level pinning after https://github.com/conda-forge/fmt-feedstock/pull/42

Issue #5339, opened 2024-01-04
Dropping CUDA 11.2


jakirkham:
Am raising this issue to discuss dropping CUDA 11.2

At this stage the bulk of conda-forge CUDA packages are built on CUDA 11.8, which supports the same hardware. Also CSPs have largely moved to CUDA 11.8+

The CUDA 11.2 Docker images that we use are [planned for deletion in May 2024]( https://gitlab.com/nvidia/container-images/cuda/-/issues/209#note_1641845842 ); whereas, the CUDA 11.8 Docker images remain available

Beginning with CUDA 12.0, standard conda-forge Docker images are used and the full CUDA Toolkit is made available with Conda packages

Given this, would propose dropping CUDA 11.2 and moving to CUDA 11.8 as the new minimum in conda-forge

Please let me know what you think

jakirkham:
cc @conda-forge/core

hmaarrfk:
I still have trouble personally running tendorflow with 11.8.

jakirkham:
Saw some things over the holidays about that, but haven't quite caught up. Is there a particular place I should start reading?

In any event this isn't an immediate thing, but beginning May 2024 maintaining CUDA 11.2 will become untenable (as the NVIDIA Docker images will be deleted)

So let's start figuring out what we need to do to sunset 11.2. Right now I have...

* TensorFlow CUDA 11.8 issues (link?)

Anything else?

hmaarrfk:
i see, i didn't know 11.2 was so old lol.

h-vetinari:
IMO we have to finish the CUDA 11.8 migration first, by which I don't mean #5340, but fixing & merging the open 11.8 PRs (or at least greatly reduce them).

> In any event this isn't an immediate thing, but beginning May 2024 maintaining CUDA 11.2 will become untenable (as the NVIDIA Docker images will be deleted)

Our existing images won't be deleted. It'll just mean we cannot rebuild them. Granted, that's a bad place to be, but it's not like everything would break from one day to the next in May.

jakirkham:
> IMO we have to finish the CUDA 11.8 migration first, by which I don't mean #5340, but fixing & merging the open 11.8 PRs (or at least greatly reduce them).

Had asked about what was still needed before closing out the migration [last meeting]( https://conda-forge.org/docs/orga/minutes/2023-12-13.html ). Didn't hear anything at that time. So we agreed to close it out

[...]

PR #5340, opened 2024-01-04
Close CUDA 11.8 migration

jakirkham:
Following up on the previous discussion on [the Dec 13 conda-forge meeting]( https://conda-forge.org/docs/orga/minutes/2023-12-13.html ), this closes out the CUDA 11.8 migration and adds it to matrix

At this point 77% of feedstocks are migrated and another 19% have an open PR to add CUDA 11.8, which accounts for 96% of feedstocks. So this is sufficiently close to done

Currently this keeps in CUDA 11.2. Though we can consider dropping this in the future as well. Have started a separate discussion in issue ( https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/5339 )


recipe/migrations/cuda120.yaml
```diff
@@ -51,6 +63,7 @@ __migrator:
       - 11.0                       # [(linux64 or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
       - 11.1                       # [(linux64 or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
       - 11.2                       # [(linux or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+      - 11.8                       # [(linux or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
```
jakirkham, line 66: When applying these changes to feedstocks with the CUDA 12.0 migrator, found that re-rendering would error due to the `ordering` not including 11.8. So have added it here

However this means feedstocks would run into this issue as well. With these changes they would need to refresh the CUDA 12.0 migrator. Alternatively we could restart this migrator to ensure the new YAML is added. Though that feels like unnecessary extra effort for maintainers. Not sure if there are better options

h-vetinari: > However this means feedstocks would run into this issue as well. With these changes they would need to refresh the CUDA 12.0 migrator.

If you update it here in the pinning repo, every feedstock that has `.ci_support/migrations/cuda120.yaml` will use that updated variant unless they set `use_local: true`. No need to restart anything.

jakirkham: If the feedstocks automatically pick up this change, great!

Had the impression that was not the case, but would be happy to be wrong

jakirkham: Yep you are right

Confirmed by testing this locally with [`conda-smithy`'s `--exclusive-config-file` flag]( https://github.com/conda-forge/conda-smithy/blob/80fd851c19fea88c040f0885d6126ab1bce8ae1d/conda_smithy/cli.py#L562-L568 )

jakirkham:
@conda-forge/core any more thoughts on this?

mariusvniekerk (approved):

kkraus14 (approved):

jakirkham:
Thanks all! 🙏

recipe/migrations/cuda120.yaml
jakirkham: Have since done a few re-renders on feedstocks with both the CUDA 11.8 & 12 migrator included without issues

For the `use_local: true` case, searched through the feedstocks and found a handful that were affected. Submitted PRs to them to refresh their CUDA 12 migrators so that re-render works. Added some background in the PRs for context

jakirkham: Since the `use_local: true` workaround was added by feedstocks working around a CUDA 12 + GCC 12 + pybind11 bug ( https://github.com/pybind/pybind11/issues/4606 ), pybind11 made a fix ( https://github.com/pybind/pybind11/commit/3414c56b6c7c521d868c9a137ca2ace2e26b5b2e ), which was released in [`pybind11` version `2.12.0`]( https://github.com/pybind/pybind11/releases/tag/v2.12.0 ). Coincidentally the same version that included the NumPy 2 fixes that we also need ( https://github.com/pybind/pybind11/commit/705efccecdf8632612dfb538008efa02c862d615 ). This newer version of `pybind11` was packaged in conda-forge earlier this year ( https://github.com/conda-forge/pybind11-feedstock/pull/94 ). So for feedstocks using `pybind11` directly, simply getting the latest version in conda-forge was sufficient for them to drop these workarounds

However the other issue we had was feedstocks that use PyTorch got `pybind11` from a vendored copy in PyTorch (not the package). This comes up when a package `#include`s PyTorch's headers. Fortunately PyTorch upgraded its version of pybind11 to 2.12.0 as part of their NumPy 2 upgrade effort ( https://github.com/pytorch/pytorch/commit/6c2f36c9845f310db8ece23c0d2e4ad6f702bc57 ). This was released in [PyTorch version 2.4.0]( https://github.com/pytorch/pytorch/releases/tag/v2.4.0 ), which was recently packaged in conda-forge ( https://github.com/conda-forge/pytorch-cpu-feedstock/pull/250 ). This allowed the remaining workarounds to be dropped

While some of these workarounds were removed organically over time, a few were still around recently. So have worked with maintainers to clean these out. At this point, I see [none of the original vendored CUDA 12 migrators]( https://github.com/search?q=org%3Aconda-forge+path%3A.ci_support%2Fmigrations%2Fcuda120.yaml+++%23+Vendor+CUDA+12+migrator+to+use+GCC+11.+++%23+This+is+needed+to+workaround+a+pybind+%2B+GCC+12+%2B+nvcc+12+bug+++%23+xref%3A+https%3A%2F%2Fgithub.com%2Fpybind%2Fpybind11%2Fissues%2F4606&type=code )

There are [other cases where the CUDA 12 migrator is vendored]( https://github.com/search?q=org%3Aconda-forge+path%3A.ci_support%2Fmigrations%2Fcuda120.yaml++use_local&type=code ). However this is usually done in packages where they build against the latest CUDA 12 and can continue to support older CUDA 12's. So this is irrelevant in this context

Mentioning here as the other context is in this thread or xref'd below. So this seemed like the natural place to include this information

h-vetinari: I think you're commenting on the wrong PR...?

jakirkham: The comment relates to both PRs, but felt more natural here given more relevant context here

Also wanted to reference it in the other PR as done in comment: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/6263#issuecomment-2323109069

Assuming that was the PR you had in mind

PR #5415, opened 2024-01-24
Add OpenSlide to pinnings and start OpenSlide 4 migrator

jakirkham:
As OpenSlide has a `run_exports` set starting with 3.4.1 ( https://github.com/conda-forge/openslide-feedstock/pull/16 ), think it makes sense to add it to the global pinnings. Starting with 3 there, but have also added a migrator for 4, which was recently added ( https://github.com/conda-forge/openslide-feedstock/pull/24 ). This will help us step from OpenSlide 3 to 4 (especially as some things like OpenSlide-Python still use 3)

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -660,6 +660,8 @@ openjpeg:
   - '2'
 openmpi:
   - 4
+openslide:
+  - 3
 openssl:
   - '3'
 orc:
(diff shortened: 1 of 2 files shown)


jakirkham:
cc @conda-forge/openslide @conda-forge/openslide-python (for awareness)

jakirkham:
@hmaarrfk can you please give this a look and let me know what you think 🙂

hmaarrfk:
Thanks!

PR #5443, opened 2024-01-29
WIP: Migrate for zlib 1.3?

h-vetinari:
There was a new zlib build release last August (a few weeks [ago](https://github.com/conda-forge/zlib-feedstock/pull/76) in conda-forge; it seems the bot didn't open a PR yet).

We've been on zlib 1.2 since the beginning of the global pinning
https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/754eda87ebfaec98adae77eebb63d7e7c0452777/recipe/conda_build_config.yaml#L195-L196
so this would probable be a relatively big migration. It's possible that migrating isn't even necessary though. While there's been no update in the [abi lab](https://abi-laboratory.pro/?view=timeline&l=zlib) for a long time, the newest `libzlib` still [contains](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Flinux-64%2Flibzlib-1.3-hd590300_0.conda) both:
```
lib/libz.so.1
lib/libz.so.1.3
```
indicating that major-level pinning the SOVER should be enough.

The [changelog](https://github.com/madler/zlib/blob/develop/ChangeLog) for zlib 1.3 also doesn't look particularly scary or ABI-relevant, though I guess the minor number was increased due to things like https://github.com/madler/zlib/issues/633, which caused a pretty substantial overhaul, see e.g. https://github.com/madler/zlib/commit/e9d5486e6635141f589e110fd789648aa08e9544

We should check if the ABI changed; if not, we could relax the pins.

CC @conda-forge/zlib @conda-forge/core

### DIFF
--- a/recipe/migrations/zlib13.yaml
+++ b/recipe/migrations/zlib13.yaml
@@ -1,0 +1,7 @@
+__migrator:
+  build_number: 1
+  kind: version
+  migration_number: 1
+zlib:
+- '1.3'
+migrator_ts: 1706542673.1618695


pitrou:
> This looks like "cosmetic" changes. Can someone try whether e.g. libarchive or python works with the new zlib without rebuilding the current binaries?

I agree that nothing significant seems to have changed in the zlib's public headers. That's probably why it's still providing a `libz.so.1` and not `libz.so.2`.

hmaarrfk:
I tried to build a few packages for zlib 1.3 but honestly, i think i ran into some rerendering issues and gave up.

omron93:
@hmaarrfk @h-vetinari Is there any progress about this and getting the zlib updated to 1.3 for fixing CVE-2023-45853?
How can I help with that? (I don't have previous experience with conda-forge building)

h-vetinari:
> How can I help with that?

You could install `zlib=1.2` and `zlib=1.3` into two different environments, and then try following these [instructions](https://github.com/lvc/abi-dumper?tab=readme-ov-file#check-for-abi-changes) from abi-dumper (which is what used to powers abi-laboratory, though unfortunately it seems the pages aren't getting updated anymore, including [zlib](https://abi-laboratory.pro/?view=timeline&l=zlib)) on `/path/to/env1_or_env2/lib/libz.so` between the two environments, and report back the results, or where you ran into problems.

omron93:
@h-vetinari Trying that now with builds from conda-forge, do you know if it's possible get builds with debug info included or striped in different package?

omron93:
For some reason the attachment was not added to previous message [compat_reports.zip](https://github.com/conda-forge/conda-forge-pinning-feedstock/files/15418195/compat_reports.zip) :-)

h-vetinari:
Thanks for running this @omron93! 🙏 

Taking a screenshot from the produced report, it indeed looks like the ABI is untouched:

image

PR #5555, opened 2024-02-22
bump re2 to 2023.09.01

h-vetinari:
This explicitly does _not_ come with a migration, because we re2 now has a SOVER, which hasn't [changed](https://github.com/google/re2/blob/2023-09-01/CMakeLists.txt#L31-L33) from 2023.06.02, and we [refactored](https://github.com/conda-forge/re2-feedstock/issues/67) the re2 outputs so that the `re2` version should just be possible to bump in the pinning _without_ a migration, whereas `libre2-11` reflects the ABI/SOVER (i.e. changes there need a migration).

This is the [first](https://github.com/conda-forge/re2-feedstock/pull/71) version bump after the refactoring, so taking time to double-check here. 

CC @conda-forge/re2

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -757,7 +757,7 @@ qtkeychain:
 rdma_core:
   - '49'
 re2:
-  - 2023.06.02
+  - 2023.09.01
 readline:
   - "8"
 rocksdb:


h-vetinari:
Ping for comments @conda-forge/re2 @isuruf - planning to merge this soonish if there are no further comments.

xhochy:
What is the motivation of bumping this? We could keep the pin as-is for quite a long time.

h-vetinari:
The motivation is that some libraries (like or-tools) need newer re2 versions. We should definitely not get stuck on old versions - if so, the refactoring would IMO have been a failure.

The idea (from my POV) is still to keep the version as up-to-date as any other library, but rebuild less often (only when the SO_VER changes). That translates into a migrator-less version bump here.

PR #5460, opened 2024-02-02
add combined migration for abseil/grpc/protobuf

h-vetinari:
Combined migration for [libprotobuf v25.2](https://github.com/conda-forge/libprotobuf-feedstock/pull/202), [grpc 1.61](https://github.com/conda-forge/grpc-cpp-feedstock/pull/348) and [abseil 20240116.0](https://github.com/conda-forge/abseil-cpp-feedstock/pull/71).

~Leaving this as a draft until all builds are green~

Closes #5413

PR #5557, opened 2024-02-22
start migration for libabseil, libgrpc & libprotobuf

h-vetinari:
Actually switch on migration added in #5460, now that the last blocker is getting solved.

Closes #5474
Closes #5492


h-vetinari:
Alright, ready to rumble!

PR #5607, opened 2024-03-06
fix c_stdlib_version zip with cdt_name

h-vetinari:
Follow-up to #5592

This tries to match the setup of the linux selectors with cdtname:
https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/595a16bb6088c5c90fc7140d15ba5b507c262245/recipe/conda_build_config.yaml#L123-L132

CC @beckermr @jakirkham @isuruf

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -13,8 +13,11 @@ c_stdlib:
   - macosx_deployment_target   # [osx]
   - vs                         # [win]
 c_stdlib_version:              # [unix]
-  - 2.12                       # [linux and x86_64]
+  - 2.12                       # [linux64 and os.environ.get("DEFAULT_LINUX_VERSION", "cos6") == "cos6"]
+  - 2.17                       # [linux64 and os.environ.get("DEFAULT_LINUX_VERSION", "cos6") == "cos7"]
   - 2.17                       # [linux and not x86_64]
+  - 2.17                       # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+  - 2.17                       # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
   - 10.9                       # [osx and x86_64]
   - 11.0                       # [osx and arm64]
 cxx_compiler:
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -164,7 +164,6 @@ zip_keys:
     - c_compiler_version        # [unix]
     - cxx_compiler_version      # [unix]
     - fortran_compiler_version  # [unix]
-    - c_stdlib                  # [unix]
     - c_stdlib_version          # [unix]
     - cdt_name                  # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
     - cuda_compiler             # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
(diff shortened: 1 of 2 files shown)


h-vetinari:
> Do we need to do this with `c_stdlib` too as it is [also zipped](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/595a16bb6088c5c90fc7140d15ba5b507c262245/recipe/conda_build_config.yaml#L164-L165)?

I don't believe so; for example, `c_compiler`, is part of the same zip as well but has no extra lines for `CF_CUDA_ENABLED` or in `cuda120.yaml`. That's because the zip itself changes based on `CF_CUDA_ENABLED`:
https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/2ecff6dac89418ae88365424707d69ca7ff1e861/recipe/conda_build_config.yaml#L159-L169

So now that we need to bump to 2.17 for the CUDA-enabled builds, I'll have to touch that the way that `c_stdlib_version` is zipped. 😑

h-vetinari:
> Do we need to do this with `c_stdlib` too as it is [also zipped](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/595a16bb6088c5c90fc7140d15ba5b507c262245/recipe/conda_build_config.yaml#L164-L165)?

So I looked at this again in more detail, and I had wrongly remembered / imagined that `c_compiler` needed to be zipped with its version. That's why it doesn't have config for `CF_CUDA_ENABLED`, and why it's not updated as part of the CUDA migrator. I think the same situation applies to `c_stdlib`, so I've removed that from the zip_keys. 

> Think we may also need to add both of these to [the CUDA 12 migrator](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/595a16bb6088c5c90fc7140d15ba5b507c262245/recipe/migrations/cuda120.yaml) as well

Consequently, I've added only `c_stdlib_version` to `cuda120.yaml`.

I was kinda expecting the CI here would check matching zip length also for CUDA builds? But in any case, I think this should be good now. I'd propose to merge it and see if it solves the ucx situation.

beckermr:
@h-vetinari @jakirkham can you all test these changes with relevant feedstocks before merging?

jakirkham:
Yes am planning to test. Just have been in meetings most of the day

h-vetinari:
AFAIU it's not possible to fully test this because the change in the zip_keys will only apply once the global pinning is updated.

beckermr:
You can run against the new copy of the config by download it locally and calling

```
conda-smithy rerender --no-check-uptodate -e conda_forge_pinning.yaml
```

h-vetinari:
OK, I can confirm that https://github.com/conda-forge/ucx-split-feedstock/pull/158 can rerender again with the changes from this PR.

beckermr (approved):
Thank you!

jakirkham:
Thanks all! 🙏

This is working smoothly now 😄

njzjz:
I got the glibc error for cuda12 builds today. It installed sysroot 2.28 instead of 2.17, while the glibc in the image is still 2.17. Finally, I found this PR and realized cuda120.yaml should be updated manually. The update for cuda120.yaml here doesn't affect the downstream feedstocks.

njzjz:
> I got the glibc error for cuda12 builds today. It installed sysroot 2.28 instead of 2.17, while the glibc in the image is still 2.17. Finally, I found this PR and realized cuda120.yaml should be updated manually. The update for cuda120.yaml here doesn't affect the downstream feedstocks.

Wait, it doesn't work...

PR #5672, opened 2024-03-24
Only zip c_stdlib_version with compilers for CUDA

mbargull:
Checklist

There's no need to put the `c_stdlib_version` (and related keys) into the compiler-version zip key.
This tight coupling is only needed for the CUDA stuff.


beckermr (approved):

h-vetinari (approved):
Fine by me, though I'd like to see it tested (like #5607; instructions therein) against some relevant feedstocks, e.g. https://github.com/conda-forge/ucx-split-feedstock.

h-vetinari:
> I'd like to see it tested (like #5607; instructions therein) against some relevant feedstocks, e.g. https://github.com/conda-forge/ucx-split-feedstock.

I did this now (both with and without https://github.com/conda-forge/conda-smithy/pull/1889), and ucx-split, cupy, abseil-cpp, numpy all worked fine.

It's important to note that we _cannot_ remove (from #5607)
https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/e29dc0a421641087716a75218da7d5936b2458ad/recipe/migrations/cuda120.yaml#L93-L94
as the CUDA 12 builds still need that for the zip to work (though the effect is somewhat surprising - not an error, but the CUDA 12 builds simply getting dropped); I experienced that because
```
conda-smithy rerender --no-check-uptodate -e conda_forge_pinning.yaml
```
ends up not pulling in the newest migrator versions, and so this happened for ucx-split & cupy.

mbargull:
FYI, AFAICT, there's only one feedstock for which the new `cdt_name` zipping would caused issues; I've address that in the feedstock at https://github.com/conda-forge/libusb-feedstock/pull/22 .
(And we only needed to add `c_stdlib_version` there without any of the compiler versions due to this PR, of course.)

PR #4086, opened 2023-02-09
petsc 3.18 migration

minrk:

PR #5708, opened 2024-04-03
merge petsc, slepc migrations

minrk:
so anything that depends on both will get upgraded at the same time

not sure if this needs to be a new migrator?

### DIFF
--- a/recipe/migrations/slepc321.yaml
+++ b/recipe/migrations/slepc321.yaml
@@ -1,8 +1,0 @@
-__migrator:
-  build_number: 1
-  commit_message: Rebuild for slepc 3.21
-  kind: version
-  migration_number: 1
-migrator_ts: 1711997957.7520285
-slepc:
-- '3.21'
--- a/recipe/migrations/petsc321.yaml
+++ b/recipe/migrations/petsc321.yaml
@@ -6,5 +6,9 @@ __migrator:
 migrator_ts: 1711978255.2022784
 petsc:
 - '3.21'
+petsc4py:
+- '3.21'
 slepc:
+- '3.21'
+slepc4py:
 - '3.21'
--- a/recipe/migrations/petsc321.yaml
+++ b/recipe/migrations/petsc321.yaml
@@ -6,3 +6,5 @@ __migrator:
 migrator_ts: 1711978255.2022784
 petsc:
 - '3.21'
+slepc:
+- '3.21'


recipe/migrations/slepc321.yaml
```diff
@@ -1,8 +0,0 @@
-__migrator:
-  build_number: 1
-  commit_message: Rebuild for slepc 3.21
-  kind: version
-  migration_number: 1
-migrator_ts: 1711997957.7520285
-slepc:
-- '3.21'
```
h-vetinari, line 8: You could just add `petsc:` here as well. While the migrations would be redundant that way, it'd have the advantage of not (effectively) restarting the slepc migration. Not sure how far it progressed and whether that's a concern, but just as an idea.

minrk: it started yesterday and hasn't gotten anywhere. There are only a few packages in it. Most things that depend on slepc also depend on petsc.

Usually we do a single migration for all of the petsc/slepc packages, like #4086

minrk:
The lack of merge here is producing broken PRs like https://github.com/conda-forge/fenics-dolfinx-feedstock/pull/70 and https://github.com/conda-forge/fenics-dolfinx-feedstock/pull/69

h-vetinari (approved):

jakirkham:
If you need to, you can restart the migration

Though defer to you whether to do that or not

h-vetinari:
> If you need to, you can restart the migration

Let's please not restart migrations defensively. The bot will pick up the new version just fine, and take all feedstocks into account that depend on one of the keys being updated.

I say this because the CUDA 12.0 migrator bumped its migration number a couple of times (didn't validate that it was necessary, but my suspicion is it wasn't), which breaks smithy pulling in the newest version for already-migrated feedstocks. This caused a bunch of feedstocks to break (e.g. as the 11.8 migration closed and the CUDA 12.0 migrator needed updates, which didn't get applied when rerendering to feedstocks that had a previous version), until someone bumped the migration number in-repo, or (more often than not) just overwrote `cuda120.yaml` with the one from the pinning.

This kind of manual overhead can be avoided if we don't needlessly restart migrations. Feel free to ping me if you're unsure in the future.

jakirkham:
Am merely stating it is an option should it be needed. We can of course discuss if Min finds he needs that option for that reason

The CUDA 12 migration needed restarts as new OSes or architectures were added. In those cases it was brought up and discussed in the core meeting first where we agreed restarting the migrator was needed otherwise we would have issues with dependencies that had not been rebuilt

Most of the pain points with the CUDA 12 migrator can be attributed to feedstocks adding `use_local`, which is due to needing to workaround pybind11 issues. This was discussed in this thread ( https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/5340#discussion_r1441268175 ) and I personally went through and updated the affected feedstocks

In any event, please do not commingle the CUDA 12 migrator with restarting in Min's case. This one should be discussed on its own merits (again only if it is needed)

h-vetinari:
> Most of the pain points with the CUDA 12 migrator can be attributed to feedstocks adding `use_local`,

That ("most") is not correct. Feedstocks with `migration_number: 1` would fail to rerender unless updated to `migration_number: 3`, completely without any `use_local`; I personally did several of those.

> In any event, please do not commingle the CUDA 12 migrator with restarting in Min's case. This one should be discussed on its own merits (again only if it is needed)

You're right that the increments for CUDA were justified by the addition of the arches and win (though even in that case, a migration bump is not imperatively necessary), but the rerendering issues that followed are more than enough reason to be very careful about migration bumps. That point still stands, and since, additionally, no feedstock gets misrendered/mismigrated by this change, there is in fact no merit to a bump here.

jakirkham:
Again let's not have this discussion here. It is orthogonal to the migration being discussed

h-vetinari:
No it isn't. The point I'm making is that a migration restart is not justified here, bolstered by an example where that lead to issues (despite being justified _there_).

jakirkham:
No one has argued that there needs to be a restart. Merely that if issues occur, it is an option (only after discussion here)

PR #5730, opened 2024-04-08
Pin snappy at minor level

xhochy:
See https://github.com/conda-forge/snappy-feedstock/issues/35

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -836,7 +836,7 @@ sdl2_ttf:
 singular:
   - 4.3.2.p8
 snappy:
-  - 1
+  - 1.1
 soapysdr:
   - '0.8'
 sox:


xhochy:
Will merge once https://github.com/conda-forge/snappy-feedstock/pull/37 is uploaded.

beckermr:
Do we want a formal migration here so that we ensure every feedstock that needs it gets a rerender?

xhochy:
> Do we want a formal migration here so that we ensure every feedstock that needs it gets a rerender?

Sorry, this comment only loaded part merge.

I will add a 1.2 migration soon. We already have a repodata patch in place.

xhochy:
> we'll have to live with the possibility that something might get an old version.

Yes, this is what we did in past cases. I will put in the migrator to 1.2 now and that should fix it. From my understanding, 1.2 is API but not ABI compatible, so we probably move the stack quite quickly.

PR #5750, opened 2024-04-11
update pin for r-base in aarch64

juiceonion:
Checklist

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -812,6 +812,7 @@ ruby:
   - 2.5
   - 2.6
 r_base:
+  - 3.6.3 # [linux and aarch64]
   - 4.1   # [win]
   - 4.2   # [not win]
   - 4.3   # [not win]


recipe/conda_build_config.yaml
```diff
@@ -812,6 +812,7 @@ ruby:
   - 2.5
   - 2.6
 r_base:
+  - 3.6.3 # [linux and aarch64]
```
martin-g, line 815: Cross-ref: https://github.com/bioconda/bioconda-recipes/pull/47149/files#r1560633610
When I recommended to add a pin for r-base I didn't know that there is such already because I was searching for `r-base` (note the `-` instead of `_`).

The versions listed below say that on Windows the build will use 4.1. On other platforms (including Linux ARM64) the builds could use 4.2 or 4.3.
Your new addition will pin it to 3.6.3 for Linux ARM64 that is older than 4.2/4.3.
Building Bioconda recipes should use 4.x.
So, there is nothing to do here. I think this PR could be closed.

mfansler, line 815:
```suggestion
```

mfansler (changes requested):
This should be closed as per @martin-g recommendation.

PR #5753, opened 2024-04-11
Add miration and global pin for c-blosc2

hmaarrfk:
Larger discussion in: https://github.com/conda-forge/c-blosc2-feedstock/issues/62#issuecomment-2049675391

TLDR:

- Previous run export was to major version
- SO changed in between 2.13 and 2.14

Patching will be made for previous builds in https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/703

Checklist

### DIFF
--- a/recipe/migrations/cblosc2214.yaml
+++ b/recipe/migrations/cblosc2214.yaml
@@ -5,4 +5,4 @@ __migrator:
   bump_number: 1
 
 c_blosc2:
-  - '2.14'
+  - '2.14.4'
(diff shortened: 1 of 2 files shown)


recipe/migrations/cblosc2214.yaml
```diff
@@ -0,0 +1,8 @@
+migrator_ts: 1712842415
+__migrator:
+  kind: version
+  migration_number: 1
+  bump_number: 1
+
+c_blosc2:
+  - '2.14'
```
hmaarrfk, line 8:
```suggestion
  - '2.14.4'
```

hmaarrfk:
While the migration should just be to 2.14.X, I want to migrate to 2.14.4 because 2.14.0-throug3 should be considered broken since they use a different SO name.

PR #5613, opened 2024-03-09
Close CUDA 12.0 migration

jakirkham:
Now that a large majority of feedstocks have been rebuilt or have an open migration PR, let's go ahead and close out this PR and add CUDA 12 to the matrix.

This adds CUDA 12.0 & CUDA 12.4. Where CUDA 12.0 represents the oldest CUDA 12 version built against and supported in conda-forge. Similarly CUDA 12.4 represents the latest CUDA version in conda-forge. This way if users are still have packages that depend on an older CUDA 12 version, packages built with CUDA 12.0 will satisfy that need. Also if users are trying to take advantage of packages built with the latest optimizations, they can install packages built with the latest CUDA Toolkit in conda-forge. We keep the latest version pinned to avoid pulling in partially rolled out CUDA Toolkit updates. Meaning we will need to go back through and bump this number when a new CUDA Toolkit becomes available.

xref: https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/5390

<hr>

Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #5799, opened 2024-04-23
Drop CUDA 11.2

h-vetinari:
The 22nd of April has passed (where we [announced](https://conda-forge.org/news/2024/03/06/dropping-cuda-112/) we'd drop CUDA 11.2). This is a variant of #5613 that drops 11.2 and incorporates the various review requests there.


jakirkham (changes requested):
Thanks Axel! 🙏

Let's stick to dropping CUDA 11.2 here

We can finish our discussion of how CUDA 12 is added in PR: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/5613

h-vetinari:
> We can finish our discussion of how CUDA 12 is added in PR: #5613

Sure we can discuss there.

> Let's stick to dropping CUDA 11.2 here

I'd prefer to do both at the same time.

h-vetinari:
> I'd prefer to do both at the same time.

It's more than a preference actually. If we don't do it at once, we'll break feedstocks that override a GCC version on a CUDA-enabled feedstock. I've recently had to do that for arrow: https://github.com/conda-forge/arrow-cpp-feedstock/commit/af240dccfdf5df1daf4419f3e8ea2ebda1169b5e

(The arrow feedstock will still need fixing afterwards, but at least my approach wouldn't break the bot from opening upgrade PRs.)

h-vetinari:
Friendly ping @jakirkham. I say we should avoid changing the lengths of the compiler-version zip without a migration, unless there is an urgent overriding cause. I don't see one here. Happy to discuss this in the core call if you prefer.

jakirkham:
Thanks for the reminder! Have been a bit swamped recently

Interesting context. Yeah let's discuss in the core call. Added to the agenda

jakirkham:
Based on discussion in the core call, it sounds like we are ok with dropping CUDA 11.2 on its own. Would it be alright if we update this PR to reflect that? Then think we can merge

h-vetinari:
> Based on discussion in the core call, it sounds like we are ok with dropping CUDA 11.2 on its own.

That's not exactly what was discussed IMO, _but_ given that we've deferred the closing of the 12.0 migration for now, I'm fine to close 11.2 separately and deal with the fall-out in arrow.

jakirkham:
Am happy to also look at Arrow if you need a hand

h-vetinari:
> Am happy to also look at Arrow if you need a hand

It's fine, thanks. I know how to do it, I just didn't want the migration bots to stop issuing PRs due to this.

jakirkham:
Completely understand

jakirkham (approved):

jakirkham:
Thanks Axel! 🙏

Have gone ahead and approved. Please feel free to merge when you are ready

PR #5857, opened 2024-05-06
Revert "Add aarch64 build for yaggo"

martin-g:
Reverts conda-forge/conda-forge-pinning-feedstock#5853

yaggo is a Bioconda recipe. There is no feedstock for it!


hmaarrfk:
hmm, we should probably lint against this kind of stuff.

PR #5893, opened 2024-05-15
close libarrow 16 migration

h-vetinari:
The only remaining [PR](https://github.com/conda-forge/turbodbc-feedstock/pull/121) is essentially ready, just waiting for merge.


recipe/conda_build_config.yaml
```diff
@@ -446,16 +446,17 @@ libarchive:
 # keep in sync with arrow_cpp (libarrow exists only from 10.x,
 # but make sure we have same length for zip as arrow_cpp)
 libarrow:
+  - 16
   - 15
   - 14
   - 13
   - 12
```
kkraus14, line 453: Should we drop this 12 as well?

h-vetinari: Yeah; it must have been a race condition between save & commit on my part...
```suggestion
```

PR #5903, opened 2024-05-17
WIP: add github_actions_labels to compiler-zip for CUDA builds

h-vetinari:
This is a draft for discussion to solve a problem in https://github.com/conda-forge/pytorch-cpu-feedstock/pull/238.

In general, for cirun-enabled feedstocks, we need to distinguish the GHA runner in the same way that we need to distinguish the docker_images (mostly CUDA vs. non-CUDA). However, that's basically impossible to pull off in a local CBC, because it would need to add to the zip containing `cuda_compiler`, which isn't possible. The only way I currently see to fix this is to add the key here, but leave it empty, for feedstocks to override.


h-vetinari:
OK, we took a simpler approach for pytorch, so this isn't necessary.

PR #6301, opened 2024-08-16
clarify run-time requirements for numpy in migrator

h-vetinari:
From discussion in https://github.com/conda-forge/numpy-feedstock/issues/324

CC @hmaarrfk

### DIFF
--- a/recipe/migrations/numpy2.yaml
+++ b/recipe/migrations/numpy2.yaml
@@ -16,7 +16,7 @@ __migrator:
     run-export on the numpy package itself. The migrator will therefore remove
     any occurrences of this.
     
-    However, you will likely still need to add the lower bound for the numpy version,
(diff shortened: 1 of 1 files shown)


jakirkham (approved):
Thanks Axel! 🙏

hmaarrfk:
The whole paragraph:

>    Additionally, we no longer need to use `{{ pin_compatible("numpy") }}` as a
>    run requirement - this has been handled for more than two years now by a
>    run-export on the numpy package itself. The migrator will therefore remove
>    any occurrences of this.

should be adjusted

h-vetinari:
I think the content of that paragraph is still correct, but I added a new one to clarify that removing `pin_compatible` does not mean that there shouldn't be run-requirements. Feel free to suggest ways to improve/shorten this.

hmaarrfk (approved):

PR #6338, opened 2024-08-25
Also add pinning for libboost_headers

hmaarrfk:
The current status of vtk only depends on this and I feel like the version is getting vendored somewhere

https://github.com/conda-forge/vtk-feedstock/blob/2c370467f4199dde94278ebf6152db72e99084c2/recipe/meta.yaml#L102

See how petpvc finds a vendoered version of boost
https://github.com/conda-forge/petpvc-feedstock/pull/11

Checklist

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -449,6 +449,8 @@ libblitz:
   - 1.0.2
 libboost_devel:
   - '1.84'
+libboost_headers:
+  - '1.84'
 libboost_python_devel:
   - '1.84'
 libcint:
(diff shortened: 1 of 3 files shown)


h-vetinari:
I'm ambivalent here. Both are OK for me -- very few packages that depend on only the headers actually break on a version upgrade (so they could add a cap until compatible), but since there _are_ some that break, that's an argument to add it to the pinning. 🤷

hmaarrfk:
the vtk maintainers have continued to emphasize that their headers are important since they use alot of templating. if you think this is OK, then merging would be appreciated!

Tobias-Fischer:
+1 for this PR. I think the breakages warrant rebuilding a few packages that wouldn't necessarily need it.

h-vetinari (approved):

PR #6424, opened 2024-09-17
migrations/osx_arm64.txt: add fluidfft and fluidsim

paugier:
I would like to have fluidsim and fluidfft added to the migrations/osx_arm64.txt. Is it correct that I need to propose this as a PR?


wolfv:
yep!

PR #6452, opened 2024-09-22
add new abseil/grpc/protobuf migration

h-vetinari:
This has been a long time coming, but we managed to unblock the protobuf situation now after upstream forced a build backend change to bazel (which itself depends on protobuf in our builds).

I'm adding the migration as paused for now to build out the baseline packages.

See also #4075, and the sketched [plan](https://github.com/conda-forge/protobuf-feedstock/pull/215#issuecomment-2360810110). There's already grpc 1.66 & libprotobuf 5.28.2, but given that this is all very fresh, I'd prefer to do smaller version jumps, and fix things as we go. Once we've got this migration out of the way, we can jump to the newest builds.

CC @conda-forge/abseil-cpp @conda-forge/grpc-cpp @conda-forge/libprotobuf @conda-forge/protobuf

### DIFF
--- a/recipe/migrations/libabseil20240722_grpc165_libprotobuf5275.yaml
+++ b/recipe/migrations/libabseil20240722_grpc165_libprotobuf5275.yaml
@@ -1,0 +1,18 @@
+__migrator:
+  build_number: 1
+  commit_message: Rebuild for libabseil 20240722, libgrp 1.65 & libprotobuf 5.27.5
+  kind: version
+  paused: true
+  migration_number: 1
+  exclude:
+    - abseil-cpp
+    - grpc-cpp
+    - libprotobuf
+    - protobuf
+libabseil:
+- 20240722
+libgrpc:
+- "1.65"
+libprotobuf:
+- 5.27.5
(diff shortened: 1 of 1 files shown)


hmaarrfk:
I'm a little concerned that I just *just* merged the `mkl=2024` migration which may have alot of overlap with packages that use protobuf.

I am also selfishly concerned with the fact that I've been stuck on pytorch's aarch build https://github.com/conda-forge/pytorch-cpu-feedstock/pull/256 for the last 3 weeks (yes many are helping, but we are still working through things).

h-vetinari:
We're not ready to start migrating (hence also why this migration is still paused in this PR). We need at least https://github.com/conda-forge/grpc-cpp-feedstock/pull/370 for example.

Happy to wait for MKL to advance a bit, a few days more or less will not change things after months of waiting. However, I don't want to wait for the likes of pytorch/tensorflow - those move at a slower speed than the rest of conda-forge, so it'd be completely fine IMO to finish your PR there as-is in however long that takes, and we worry about the migration for pytorch afterwards.

PR #6484, opened 2024-09-29
Add UCC to global pinning

leofang:
cc @conda-forge/ucc @conda-forge/openmpi

Checklist

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -885,6 +885,8 @@ tk:
   - 8.6                # [not ppc64le]
 tiledb:
   - '2.26'
+ucc:
+  - 1
 ucx:
   - '1.17'
 uhd:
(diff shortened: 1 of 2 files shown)


hmaarrfk:
Your run export is for the major version only.

can we just add a pin of "1" to the global pinning?
```
    - {{ pin_subpackage("ucc") }}
```

https://github.com/conda-forge/ucc-feedstock/blob/main/recipe/meta.yaml

leofang:
good catch, done

hmaarrfk:
The earliest version of the package was 1.2.

so any 1.2 build should be compatibile even with 1.X.

so you can just directly add it to https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/conda_build_config.yaml#L888

```
ucc:
   - 1
```

leofang:
Sure, if touching `conda_build_config.yaml` directly is better. Done.

hmaarrfk:
> Sure, if touching conda_build_config.yaml directly is better. Done.

In these cases where you are being proactive, its just simpler and avoids rebuilds!

leofang:
Thanks, Mark!

PR #6533, opened 2024-10-12
delete obsoleted grpc/protobuf migration

h-vetinari:
This became overtaken/obsoleted by https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/6452


xhochy:
Shouldn't we update the pin?

h-vetinari:
> Shouldn't we update the pin?

We never migrated anything for this. It was merged as paused so I could use it to more easily rerender the initial set of baseline feedstocks (i.e. without having to manually create overrides in CBC everywhere that would later have to be deleted again).

recipe/migrations/libgrpc163_libprotobuf5261.yaml
```diff
@@ -1,11 +0,0 @@
-__migrator:
-  build_number: 1
-  commit_message: Rebuild for libgrp 1.63 & libprotobuf 5.26.1
-  kind: version
-  paused: true
```
h-vetinari, line 5: see here; it was never unpaused, because we never managed to build `protobuf` for 5.26

xhochy (approved):
Thanks! I thought we started this already.

PR #6548, opened 2024-10-15
[RFC]: Add `os_version` option to use `alma8`

jakirkham:
Provide support for Alma 8 images based on GLIBC 2.28 with the updated CDTs.

* Checks `DEFAULT_LINUX_VERSION` for `cos7` or `alma8`
* Extends `docker_image`, `cdt_name`, and `c_stdlib_version`
* Includes selectors to support the different combinations
* For feedstocks not using `os_version`, this is a no-op

Could be extended to CUDA 11.8 builds as well, but would require an additional Docker image, which Axel kindly put together in PR ( https://github.com/conda-forge/docker-images/pull/291 ).

Also there is an overly strict check in CUDA cross-compilation support (overly restrictive on CUDA 12). Relaxing that in PR: https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/363

Some sample re-renders:

* https://github.com/conda-forge/zlib-feedstock/pull/86
* https://github.com/conda-forge/nccl-feedstock/pull/119

xref: https://github.com/conda-forge/conda-forge.github.io/issues/1941

<hr>

Checklist


carterbox:
@carterbox, please review

h-vetinari:
> Could be extended to CUDA 11.8 builds as well, but would require building additional Docker images first

The CUDA 11.8 images are already based on alma8, last time I checked.

recipe/conda_build_config.yaml
```diff
@@ -125,17 +125,24 @@ channel_sources:
 channel_targets:
   - conda-forge main
 
-cdt_name:  # [linux]
-  - cos7   # [linux]
+cdt_name:       # [linux]
+  - cos7        # [linux and os.environ.get("DEFAULT_LINUX_VERSION", "cos7") == "cos7"]
+  - conda_2_28  # [linux and os.environ.get("DEFAULT_LINUX_VERSION", "cos7") == "conda_2_28"]
```
h-vetinari, line 130: The new CDTs do not contain the image version anymore, see [here](https://github.com/conda-forge/conda-forge.github.io/blob/5e57e30a11e91dc127aab1543b2cffe8ed8b3290/community/minutes/2024-05-15.md?plain=1#L74-L77) (as pointed out in https://github.com/conda-forge/cdt-builds/pull/70).
```suggestion
cdt_name:       # [linux]
  - cos7        # [linux and os.environ.get("DEFAULT_LINUX_VERSION", "cos7") == "cos7"]
  - conda       # [linux and os.environ.get("DEFAULT_LINUX_VERSION", "cos7") == "alma8"]
```

h-vetinari (commented):
I think that for `os_version` (which is what feeds into `DEFAULT_LINUX_VERSION`), we should be using `alma8`. The `cdt_name` will eventually go away, and does not contain the os/image/glibc version.

carterbox:
Use `conda-smithy -e` to test using this pinning on a feedstock.

h-vetinari:
> The CUDA 11.8 images are already based on alma8, last time I checked (as in: they contain glibc 2.28).

That was only partially correct (for aarch/ppc), here's the fix for x64: https://github.com/conda-forge/docker-images/pull/291

Also, I think we're ready to do the move outright, see #6626. I also put this on the agenda for Thursday.

carterbox:
I tested these changes by trying to rerender the nvpl-blas feedstock.

1. Set the environment variable `DEFAULT_LINUX_VERSION=conda_2_28`
2. Rerender with `-e` pointing to this config

Questions:

This causes the docker image and cdt_name to change, but not the c_stdlib_version to change. Should it? Why would you want to build with glibc 2.17 by default in a container that has glibc 2.28?

How do switch to the conda_2_28 cdt without setting and environment variable? Can I do it by adding a static configuration file to my recipe?

carterbox:
xref: https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/6629

jakirkham:
> > The CUDA 11.8 images are already based on alma8, last time I checked (as in: they contain glibc 2.28).
> 
> That was only partially correct (for aarch/ppc), here's the fix for x64: [conda-forge/docker-images#291](https://github.com/conda-forge/docker-images/pull/291)

Thanks Axel! 🙏

> Also, I think we're ready to do the move outright, see #6626. I also put this on the agenda for Thursday.

It sounds like we still want some optionality here (`os_version` / `DEFAULT_LINUX_VERSION`), which this has.

recipe/conda_build_config.yaml
jakirkham: Thanks Axel! 🙏

Addressed in recent commits. Please take another look

jakirkham:
Thanks for testing Daniel! 🙏

Have pushed some more changes

> This causes the docker image and cdt_name to change, but not the c_stdlib_version to change. Should it? Why would you want to build with glibc 2.17 by default in a container that has glibc 2.28?

Have now added `c_stdlib_version`

One case where this might be useful is if we decided to bump all images to Alma 8. We do a pretty good job of siloing things with the compiler's sysroot, CDTs, etc. that this should largely be ok to do. This is part of what we discussed today

> How do switch to the conda_2_28 cdt without setting and environment variable? Can I do it by adding a static configuration file to my recipe?

It shouldn't be needed to set the environment variable. All that should be needed is setting `os_version` in `conda-forge.yml`. Have included some examples in the OP

jakirkham:
Marking as ready for review as this is now functional. Though leaving RFC in as it would be good to hear from more folks on these changes

cc @conda-forge/core

h-vetinari (commented):
I'd be tempted to just drop cos7 as a choice and only offer alma8/alma9 going forward (which we can similarly add here after https://github.com/conda-forge/docker-images/pull/290). I strongly doubt that there are many cases where going back to an older image is required for the build, and even then, it remains possible via overriding the `docker_image` (plus other relevant quantities for the zip) in the CBC.

If we decide to keep the option for cos7, then I'd still drop the choice for the CUDA 11.8 images, where this is a reality already for aarch/ppc (on UBI 8, see [here](https://github.com/conda-forge/docker-images/pull/291)), and so pretending it's cos7 is not helpful (many feedstocks have run into the glibc requirements on aarch especially, here's [one](https://github.com/conda-forge/libcublas-feedstock/issues/13) example)

recipe/conda_build_config.yaml
```diff
@@ -15,9 +15,11 @@ c_stdlib:
 m2w64_c_stdlib:                # [win]
   - m2w64-toolchain            # [win]
 c_stdlib_version:              # [unix]
-  - 2.17                       # [linux]
-  - 2.17                       # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+  - 2.17                       # [linux and os.environ.get("DEFAULT_LINUX_VERSION", "cos7") == "cos7"]
+  - 2.28                       # [linux and os.environ.get("DEFAULT_LINUX_VERSION", "cos7") == "alma8"]
   - 2.17                       # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+  - 2.17                       # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True" and os.environ.get("DEFAULT_LINUX_VERSION", "cos7") == "cos7"]
+  - 2.28                       # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True" and os.environ.get("DEFAULT_LINUX_VERSION", "cos7") == "alma8"]
```
h-vetinari, line 22: I'm pretty sure we don't want to touch the stdlib version here; for example, the newer image might just be required to satisfy the `__glibc` constraint of a test dependency, but doesn't necessarily mean that the package itself needs a newer glibc and run-constraint.

Despite the monster-zip it remains possible to override _one_ component locally - e.g. `c_stdlib_version`, and then set `os_version` for the image. I'd also prefer the choice of `c_stdlib_version` to be more visible than something implicitly implied by a knob tucked away in `conda-forge.yml`.

recipe/conda_build_config.yaml
```diff
@@ -146,14 +155,23 @@ docker_image:                                   # [os.environ.get("BUILD_PLATFOR
...
   - quay.io/condaforge/linux-anvil-cuda:11.8              # [aarch64 and os.environ.get("CF_CUDA_ENABLED", "False") == "True" and os.environ.get("BUILD_PLATFORM") == "linux-64"]
 
-  # CUDA 12
-  - quay.io/condaforge/linux-anvil-cos7-x86_64            # [linux64 and os.environ.get("CF_CUDA_ENABLED", "False") == "True" and os.environ.get("BUILD_PLATFORM") == "linux-64"]
+  # CUDA 12 on CentOS 7
+  - quay.io/condaforge/linux-anvil-cos7-x86_64            # [linux64 and os.environ.get("CF_CUDA_ENABLED", "False") == "True" and os.environ.get("BUILD_PLATFORM") == "linux-64" and os.environ.get("DEFAULT_LINUX_VERSION", "cos7") == "cos7"]
+  # case: native compilation (build == target)
+  - quay.io/condaforge/linux-anvil-ppc64le                # [ppc64le and os.environ.get("CF_CUDA_ENABLED", "False") == "True" and os.environ.get("BUILD_PLATFORM") == "linux-ppc64le" and os.environ.get("DEFAULT_LINUX_VERSION", "cos7") == "cos7"]
+  - quay.io/condaforge/linux-anvil-aarch64                # [aarch64 and os.environ.get("CF_CUDA_ENABLED", "False") == "True" and os.environ.get("BUILD_PLATFORM") == "linux-aarch64" and os.environ.get("DEFAULT_LINUX_VERSION", "cos7") == "cos7"]
```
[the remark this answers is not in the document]
h-vetinari: Nevermind, I confused myself on this one - these are the CUDA 12 builds, where cos7 is still a possible choice (well, except aarch in most cases).

recipe/conda_build_config.yaml
h-vetinari: > I'm pretty sure we don't want to touch the stdlib version here

OK, I've tested this hypothesis (already before commenting but) now also using a local checkout of #6626 - here's the resulting [branch](https://github.com/h-vetinari/faiss-split-feedstock/commits/pinning_test).

Basically, this works as intended if we also take the `c_stdlib_version` out of the CUDA zip, which I think is possible now. That would be the ideal scenario for me - the `c_stdlib_version` can be independently set (in CBC), from the image (in `conda-forge.yml`, or in CBC if the full zip is overridden).

carterbox:
```yaml
os_version:
  linux_aarch64: alma8
```
I added this section to my conda-forge.yml. This causes the docker image to be alma8 and the stdlib version to become 2.28.

I guess I understand @h-vetinari's point about being able to choose stdlib separately from os_version. You may want to build against an *older* stdlib version than what is in the container to improve compatibility. However, in that case, the default container glibc should always be strictly >= the latest version of glibc that we offer. In other words, we should have migrated to an OS that ships `glibc >= 2.34` already because that is available on the channel i.e. Alma 9.

carterbox:
I think we can close this one since we are going with #6626 instead.

PR #6618, opened 2024-10-26
exclude r-packages from flang migration

h-vetinari:
My understanding would be that we don't want to switch the r-packages from `m2w64_fortran` to flang, because those R-packages overwhelmingly use a consistent GCC stack (and a lot of effort went into making them ucrt-compatible).

This list was generated by taking the children of `r-base` on the [migrator status page](https://conda-forge.org/status/migration/?name=flang19)

CC @conda-forge/r @isuruf


h-vetinari:
This was agreed upon in the core call today.

PR #6643, opened 2024-11-04
Bump cuda to 12.6

hmaarrfk:
Closes https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/6630#issuecomment-2452404940

Checklist

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -17,7 +17,7 @@ m2w64_c_stdlib:                # [win]
 c_stdlib_version:              # [unix]
   - 2.17                       # [linux]
   - 2.17                       # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
-  - 2.17                       # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+  - 2.28                       # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
   - 10.13                      # [osx and x86_64]
   - 11.0                       # [osx and arm64]
 cxx_compiler:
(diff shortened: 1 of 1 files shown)


recipe/conda_build_config.yaml
```diff
@@ -17,7 +17,7 @@ m2w64_c_stdlib:                # [win]
 c_stdlib_version:              # [unix]
   - 2.17                       # [linux]
   - 2.17                       # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
-  - 2.17                       # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+  - 2.28                       # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
```
h-vetinari, line 20: The CUDA compiler version shouldn't have anything to do with the sysroot we target. Needing newer images to run some of the existing CUDA 12 build infrastructure is solved by bumping the images, not the sysroot.

h-vetinari (commented):
I would still like to migrate this first; some binaries changed in the early days of the 12.x conda-forge roll-out, but more importantly, many recipes hard-code the `cuda_compiler_version` in various places (e.g. to determine the set of GPU arches in `build.sh`), and we shouldn't just break all those recipes IMO.

Finally, we should double-check the GCC upper bound for CUDA 12.6 - we can probably go to GCC 13, matching our non-CUDA stack.

hmaarrfk:
I'm pretty bad at writing these zipped migrators and i guess I don't understand why the sysroots need to be updated so I'll let somebody else draft somethung

recipe/conda_build_config.yaml
hmaarrfk: Is this the migrator you have in mind? 

I feel like it cna likely be simplified now:

https://github.com/conda-forge/cupy-feedstock/blob/main/.ci_support/migrations/cuda12x.yaml

PR #6687, opened 2024-11-13
DEMO: consolidated image names in pinning

h-vetinari:
This is a demo for how the pinning would look like under the proposal from https://github.com/conda-forge/docker-images/issues/293

Based on #6626, so the only relevant commits are bd5223109cb32a2d0d534605928f0f2a44f24509 and 0e54e8ea2a2188452f385a983befc948edc4ae29


h-vetinari:
Changes here have now been incorporated into #6626

PR #6688, opened 2024-11-13
Provide a linter hint for xorg-proto packages

hmaarrfk:
@pkgw I think this is not the ideal way, but we have been using the newer proto on some new recipes and I am noticing clobbering in a few environments. I think it would be fine to use this in lieu of a more complicated migration

Let me know what you think

Checklist

### DIFF
--- a/recipe/linter_hints/hints.toml
+++ b/recipe/linter_hints/hints.toml
@@ -91,3 +91,7 @@ astropy = """\
     Recipes should usually depend on `astropy-base` as opposed to \
     `astropy`. `astropy-base` only has the required dependancies whereas \
     `astropy` now has all optional dependancies as well."""
+
+xorg_xextproto = """The use of `xorg-xextproto` is deprecated. \
+Please use `xorg-xorgproto` instead. It contains the latest version of the proto \
+packages and bundles them all together."""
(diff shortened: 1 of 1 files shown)


recipe/linter_hints/hints.toml
```diff
@@ -26,7 +26,7 @@ matplotlib = """\
 pathlib = """\
     As of Python 3.4+, `pathlib` is part of the standard library. \
     Recipes should drop the `pathlib` dependency as it is sunset.
-    
+
```
h-vetinari, line 29: Do not delete this whitespace, it's necessary for correct formatting of the message in a list.

hmaarrfk: thanks!

hmaarrfk:
Kind ping @pkgw

pkgw:
@hmaarrfk Sorry for being very slow to respond here. This seems like a fine start. There are of course many many other proto packages that could also be mentioned here, but Xext seems like one of the more common offenders. Maybe it would be good to also have a record for `xorg-xproto`, mainly because it seems like that one might be confusing for people due to the naming similarity.

hmaarrfk:
awesome. I just duplicated the message for both, and we'll see what happens over time.

jakirkham:
Started a discussion about a migrator for this case: https://github.com/regro/cf-scripts/issues/3360

jakirkham:
Expanded this to include all `xorg-*protos` in PR: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/6816

Please let me know what you think 🙂

PR #6741, opened 2024-11-26
add pinnings for separate suitesparse libraries

minrk:
pending https://github.com/conda-forge/suitesparse-feedstock/pull/98

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -454,6 +454,8 @@ libabseil_static:
   - '20220623.0'
 libaec:
   - '1'
+libamd:
+  - '3'
 libarchive:
   - '3.7'
 libarrow:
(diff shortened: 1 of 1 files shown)


recipe/conda_build_config.yaml
```diff
@@ -478,12 +480,24 @@ libboost_headers:
...
+libcamd:
+  - '3'
 libcint:
   - '6.1'
+libccolamd:
+  - '3'
+libcholmod:
+  - '5'
+libcolamd:
+  - '3'
```
h-vetinari, line 494: I'm of two minds if we don't want to namespace these as coming from suitesparse somehow. I know that suitesparse more or less independent projects (hence "suite"), but we're not going to get these from anywhere else realistically? No strong opinions, just wondering...

minrk: I wondered that, too! Lacking other reasons, I went with using the same names as ubuntu (`suitesparse-` is on mongoose, but not the others, I think because there are other pre-existing projects with the same name).

minrk:
The split has landed, so this can be merged, I think

PR #6758, opened 2024-11-29
Dual migration for HDF5 1.14.4 and 1.14.3

hmaarrfk:
See https://github.com/conda-forge/hdf5-feedstock/issues/240

Is this enough to retrigger the builds for the feedstocks that have already been migrated?

Checklist

### DIFF
--- a/recipe/migrations/hdf51144.yaml
+++ b/recipe/migrations/hdf51144.yaml
@@ -1,8 +1,9 @@
 __migrator:
   build_number: 1
-  commit_message: Rebuild for hdf5 1.14.4
+  commit_message: Rebuild for hdf5 1.14.4 (and keep 1.14.3)
   kind: version
-  migration_number: 1
+  migration_number: 2
 hdf5:
+- 1.14.3
 - 1.14.4
 migrator_ts: 1727986901.81392


hmaarrfk:
cc: @h-vetinari i know that you mentioned that you thought that upstream should revert their introduced bug ASAP, but in the meantime, I think this will help us at conda-forge.

h-vetinari (commented):
> cc: @h-vetinari i know that you mentioned that you thought that upstream should revert their introduced bug ASAP, but in the meantime, I think this will help us at conda-forge.

I do think that the only sane regression handling here is to revert the breakage before people start relying on the new behaviour (these are standard kernel regression rules, for example), and then introduce a more comprehensive fix on top. However, given that we have close to 0 control over how upstream deals with this, I'm OK with building for both versions.

I'd also be OK to simply drop 1.14.4, as it's essentially broken for a large part of the ecosystem.

hmaarrfk:
I feel like some people have asked for both. 

I agree with you that the breakage is large enough to warrant reverting. But I can’t control upstream. I could skip 1.14.4 for windows. What do you think?

h-vetinari:
As I said

> I'm OK with building for both versions.

I don't think it's worth dropping windows for 1.14.4 (though that's a choice that individual feedstocks can make?) - but no strong feelings and happy to be convinced otherwise

PR #6816, opened 2024-12-09
Hint for all other `xorg-*proto`s

jakirkham:
Standardize and expand the `xorg-*proto` hints from PR: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/6688

Covers all `xorg-*proto`s (excluding `xorg-xorgproto`) with similar hints and recommends they be replaced with `xorg-xorgproto`

<hr>

Checklist


hmaarrfk (approved):

recipe/linter_hints/hints.toml
```diff
@@ -91,9 +91,133 @@ astropy = """\
...
+    Please use `xorg-xorgproto` instead. It contains the latest versions of all the proto \
+    packages and bundles them all together."""
+xorg_windowswmproto = """\
+    The use of `xorg-windowswmproto` is deprecated. \
+    Please use `xorg-xorgproto` instead. It contains the latest versions of all the proto \
+    packages and bundles them all together."""
+xorg_xcb_proto = """\
+    The use of `xorg-xcb-proto` is deprecated. \
+    Please use `xorg-xorgproto` instead. It contains the latest versions of all the proto \
+    packages and bundles them all together."""
```
pkgw, line 183: The XCB protocol files are an exception: they are the only package that has *not* been merged into the generic xorgproto package. So this specific lint should be removed.

Due to the above, this second item is no longer relevant, but also: there is a legacy `xorg-xcb-proto` feedstock, but the proper name for this package is just `xcb-proto`: we had to change the name to maintain compatibility with Anaconda's `defaults`, back in the day. The `xorg-xcb-proto` feedstock should be archived.

pkgw (changes requested):
Thanks for going to the trouble for the full solution, here. One comment below.

recipe/linter_hints/hints.toml
pkgw: (E.g., check out: https://www.x.org/releases/individual/proto/?C=M;O=D )

jakirkham, line 183:
```suggestion
```

recipe/linter_hints/hints.toml
```diff
@@ -91,9 +91,133 @@ astropy = """\
     Recipes should usually depend on `astropy-base` as opposed to \
     `astropy`. `astropy-base` only has the required dependancies whereas \
     `astropy` now has all optional dependancies as well."""
-xorg_xextproto = """The use of `xorg-xextproto` and `xorg-xproto` are deprecated. \
-Please use `xorg-xorgproto` instead. It contains the latest version of allthe proto \
-packages and bundles them all together."""
-xorg_xproto = """The use of `xorg-xextproto` and `xorg-xproto` are deprecated. \
-Please use `xorg-xorgproto` instead. It contains the latest version of all the proto \
-packages and bundles them all together."""
+
```
jakirkham, line 94:
```suggestion

xorg_xcb_proto = """\
    The use of `xorg-xcb-proto` is deprecated. \
    Please use `xcb-proto` instead."""

```

recipe/linter_hints/hints.toml
jakirkham: Thanks Peter! 🙏 

Have tried to propose some changes based on your feedback. Please let me know whether those need make sense or need changes

jakirkham (commented):
Thanks Peter! 🙏

Would these proposed changes capture what you are looking for?

recipe/linter_hints/hints.toml
pkgw: Yeah, what you propose looks good. *Hopefully* there isn't anything that actually uses the `xorg-xcb-proto` package, since we realized that it needed to be renamed almost right off the bat, but it doesn't hurt to add a lint for it.

jakirkham: Great, have gone ahead and committed it. Please let me know if there is anything else

pkgw (approved):

jakirkham:
Thank you both! 🙏

Hoping this helps more people understand the changes needed and what to do next 🙂

h-vetinari:
Thanks all!

> Hoping this helps more people understand the changes needed and what to do next 🙂

I think the one case we don't cover yet (and which is hard/impossible to do with the hints) is those cases where the various `*proto` packages were only present transitively. I ran into such a situation in https://github.com/conda-forge/cairo-feedstock/pull/79, for example. Nothing to do for now, just noting this for completeness/visibility.

PR #6821, opened 2024-12-10
Add hint about new `xz` outputs

### LINKED ISSUE #6790: Add linter hint to move from `xz` to `liblzma-devel` in `host`/`xz*-tools` in `build`
Recently `xz` implemented a library specific output called `liblzma-devel` and different `xz*-tools` packages (with the CLI). To raise awareness, it would be nice to have a linter hint to suggest using `liblzma-devel` in `host` and the appropriate `xz*-tools` package in `build` instead. 

xref: https://github.com/conda-forge/xz-feedstock/pull/42

xhochy:
Fixes #6790

### DIFF
--- a/recipe/linter_hints/hints.toml
+++ b/recipe/linter_hints/hints.toml
@@ -97,3 +97,7 @@ packages and bundles them all together."""
 xorg_xproto = """The use of `xorg-xextproto` and `xorg-xproto` are deprecated. \
 Please use `xorg-xorgproto` instead. It contains the latest version of all the proto \
 packages and bundles them all together."""
+xz = """The use of `xz` is deprecated. \
+If you want to link to it, use the new `liblzma-devel` package. \
+For the command line utilities, you can use `xz-tools` for `xz` and `unxz` or \
+`xz-gpl-tools` for the xz-supporting CLIs that are forks of `cat` and `grep`."""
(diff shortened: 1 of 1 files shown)


recipe/linter_hints/hints.toml
```diff
@@ -97,3 +97,7 @@ packages and bundles them all together."""
 xorg_xproto = """The use of `xorg-xextproto` and `xorg-xproto` are deprecated. \
 Please use `xorg-xorgproto` instead. It contains the latest version of all the proto \
 packages and bundles them all together."""
+xz = """The use of `xz` is deprecated. \
+If you want to link to it, use the new `liblzma-devel` package. \
+For the command line utilities, you can use `xz-tools` for `xz` and `unxz` or \
+`xz-gpl-tools` for the xz-supporting CLIs that are forks of `cat` and `grep`."""
```
jakirkham, line 103: Would suggest rewording this a bit to make it a bit easier on readers to see the `if...then` logic here

Also it looks like [`cat` is in `xz-tools`]( https://github.com/conda-forge/xz-feedstock/blob/ac52d61fface9ce6cd6f9c5c410ca22ff64621b1/recipe/meta.yaml#L103 ) so dropped it from the GNU tools line

```suggestion
If your package links to `liblzma`, use the new `liblzma-devel` package. \
For XZ's command line utilities (like `xz` and `unxz`), use the `xz-tools`package. \
For the xz-enabled GNU tools (like `grep`), use `xz-gpl-tools`. \
Note: Command line utilities should go in `requirements/build`."""
```

jakirkham (commented):
Thanks Uwe! 🙏

This generally looks reasonable

Suggested some revisions to help clarify the `if` early and follow with `then`

Hoping that helps maintainers pick these lints up and act on them

PR #6849, opened 2024-12-19
bump to GCC 13 on CUDA 12.6

### LINKED ISSUE #6736: Bump GCC pins for CUDA 12.6
As a follow-up to #6720 - I just realised that bumping to 12.6 means we should have updated GCC pins as well. The upstream [docs](https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#host-compiler-support-policy) (found the reference in https://github.com/conda-forge/nvcc-feedstock/issues/51) say that nvcc compatible with "GCC 6.x - 13.2". We're on 13.3 though.

Is the minor-level bound really enforced? I don't remember that in previous nvcc constraints for gcc (which were major-only AFAIR).

_Originally posted by @h-vetinari in https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/6720#issuecomment-2495839372_

h-vetinari:
Closes #6736

Matches our default compiler pins on linux (c.f. #4890) where possible (i.e. on CUDA 12.6; on linux-ppc64le, we're stuck on CUDA 12.4, which does not support GCC 13).

Tested to work, as described in #6736.


h-vetinari:
CC @conda-forge/cuda

bdice (approved):
The version support is safe, in my understanding (13 for CUDA 12.6, 12 for CUDA 12.4).

kkraus14 (approved):

jakirkham:
Thanks all! 🙏

PR #6917, opened 2025-01-12
Reduce cuda complexity by dropping 11.8

hmaarrfk:
For years we have struggled in maintainning the multiple cuda versions.

Personally, I feel that Nvidia has done tremendous work in ameliorating the situation here at Conda-forge.

However, I'm hoping to shed a little of bit of mental weight from cuda 11.8.

IMO the main projects that utilize it:
* pytorch - https://github.com/conda-forge/pytorch-cpu-feedstock/blob/main/recipe/meta.yaml#L58
* tensorflow - https://github.com/conda-forge/tensorflow-feedstock/blob/main/recipe/meta.yaml#L60
* jax - https://github.com/conda-forge/jaxlib-feedstock/blob/main/recipe/meta.yaml#L27
* onnxruntime https://github.com/conda-forge/onnxruntime-feedstock/blob/main/recipe/meta.yaml#L48

have all dropped CUDA 11.8 for lack of maintainer interest.

I think it is time for us to drop it from the global pinnings

Checklist

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -6,7 +6,6 @@ c_compiler:
 c_compiler_version:            # [unix]
   - 13                         # [linux]
   - 18                         # [osx]
-  - 11                         # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
   # CUDA 12.4 on PPC does not support GCC 13
   - 12                         # [linux and ppc64le and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
   - 13                         # [linux and not ppc64le and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
@@ -28,7 +27,6 @@ cxx_compiler:
 cxx_compiler_version:          # [unix]
   - 13                         # [linux]
   - 18                         # [osx]
-  - 11                         # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
   - 12                         # [linux and ppc64le and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
   - 13                         # [linux and not ppc64le and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
 llvm_openmp:                   # [osx]
@@ -41,7 +39,6 @@ fortran_compiler_version:      # [unix or win64]
   - 13                         # [linux]
   - 13                         # [osx]
   - 5                          # [win64]
-  - 11                         # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
   - 12                         # [linux and ppc64le and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
   - 13                         # [linux and not ppc64le and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
 m2w64_c_compiler:              # [win]
@@ -51,18 +48,13 @@ m2w64_cxx_compiler:            # [win]
 m2w64_fortran_compiler:        # [win]
   - m2w64-toolchain            # [win]
 
-cuda_compiler:
-  - None
-  - nvcc                       # [(linux or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
-  - cuda-nvcc                  # [(linux or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
 cuda_compiler_version:
   - None
-  - 11.8                       # [(linux or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
-  - 12.4                       # [linux and ppc64le and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
-  - 12.6                       # [((linux and not ppc64le) or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+  - 12                         # [(linux or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+
 cuda_compiler_version_min:
   - None                       # [osx]
-  - 11.8                       # [linux or win64]
+  - 12                         # [linux or win64]
 
 arm_variant_type:              # [aarch64]
   - sbsa                       # [aarch64]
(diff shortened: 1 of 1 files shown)


recipe/conda_build_config.yaml
```diff
@@ -51,18 +48,13 @@ m2w64_cxx_compiler:            # [win]
 m2w64_fortran_compiler:        # [win]
   - m2w64-toolchain            # [win]
 
-cuda_compiler:
-  - None
-  - nvcc                       # [(linux or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
-  - cuda-nvcc                  # [(linux or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
```
hmaarrfk, line 57: do we really need to use this key? could we have users simply depend on cuda-nvcc? I likely don't understand the downstream implications.

h-vetinari: Users shouldn't depend on `cuda-nvcc` but continue to depend on `{{ compiler("cuda") }}`. For that, we need to set `cuda_compiler` in the pinning, also in a 12+ world

recipe/conda_build_config.yaml
```diff
@@ -177,12 +159,8 @@ zip_keys:
     - c_compiler_version        # [unix]
     - cxx_compiler_version      # [unix]
     - fortran_compiler_version  # [unix]
-    - cuda_compiler             # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
     - cuda_compiler_version     # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
     - docker_image              # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True" and os.environ.get("BUILD_PLATFORM", "").startswith("linux-")]
-  -                             # [win64 and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
-    - cuda_compiler             # [win64 and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
-    - cuda_compiler_version     # [win64 and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
```
h-vetinari, line 185: This will thus continue to be necessary

recipe/conda_build_config.yaml
```diff
@@ -51,18 +48,13 @@ m2w64_cxx_compiler:            # [win]
...
-cuda_compiler:
-  - None
-  - nvcc                       # [(linux or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
-  - cuda-nvcc                  # [(linux or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
 cuda_compiler_version:
   - None
-  - 11.8                       # [(linux or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
-  - 12.4                       # [linux and ppc64le and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
-  - 12.6                       # [((linux and not ppc64le) or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+  - 12                         # [(linux or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
```
h-vetinari, line 53: I don't want to switch to free-floating minor for 12.x in an unrelated PR. If you think that's a good idea, please open an issue and we can discuss it there.

h-vetinari (commented):
Neither for nor against the 11.8 removal, but this PR is not yet correct for its stated goal.

jakirkham:
Thanks for the kind words Mark! 🙏

Am glad to hear the CUDA 12 packages are serving the community well 😊

Can understand individual feedstock maintainers taking the step of shedding maintenance burden. And they certainly should feel free to `skip` versions they no longer can or want to support

However when it comes to the global matrix in conda-forge, we have been taking the approach of keeping a broad range of versions. For example with Python we have been keeping all versions that have not reached EOL

By the same measure we have kept around CUDA 11 (specifically 11.8). We certainly have tried to cutdown on the burden by dropping old CUDAs ( https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/1708 ), moving to CUDA 11.8 ( https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/4834 ), and dropping anything previous ( https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/5799 ). Not to mention the significant collective effort put into migrating feedstocks. All this to ensure only one version of CUDA (the latest one) is around

We do this in part because [CUDA 11.8 is the last version that still supports Kepler]( https://docs.nvidia.com/cuda/archive/11.8.0/cuda-toolkit-release-notes/index.html#title-new-cuda-libraries ). In [CUDA 12.0, Kepler support was removed]( https://docs.nvidia.com/cuda/archive/12.0.0/cuda-toolkit-release-notes/index.html#cuda-libraries ). Once we drop CUDA 11.8, those users will no longer have support

NVIDIA continues to publish its own libraries (like CUDA Python and RAPIDS with CUDA 11 support). Also a number of packages in conda-forge either from NVIDIA or maintained in part by NVIDIA, ship with CUDA 11 support, which NVIDIA helps maintain

Looking through conda-forge, [there are a number of feedstocks that continue to support CUDA 11]( https://github.com/search?q=org%3Aconda-forge+path%3A%2F%5E%5C.ci_support%5C%2F%2F+-path%3A%2F%5E%5C.ci_support%5C%2Fmigrations%5C%2F%2F+cuda+11.8&type=code ). So it seems probable that others still value providing support for CUDA 11 too

While I understand why individual maintainers would like to drop CUDA 11 from their feedstocks (and they certainly should feel free to using `skip`s). In terms of many folks still maintaining CUDA 11, it is much easier to do so with it in the global matrix

Would appreciate hearing a bit more how this issue is coming up for you. Perhaps we can find another acceptable solution

recipe/conda_build_config.yaml
jakirkham: Right we also discussed this previously. The reason we pin this here is to avoid picking up incomplete CUDA Toolkit upgrades

We do plan to bump this version whenever the next CUDA Toolkit is fully available on conda-forge

hmaarrfk:
Thanks for the thorough explanation

Issue #7029, opened 2025-02-13
Error on migration with `.yml` extension; should be `.yaml`


jakirkham:
From @isuruf:

> We can add an error at https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/run_test.py

xref: https://github.com/regro/cf-scripts/issues/3704#issuecomment-2656827766

PR #7036, opened 2025-02-14
raise an error when *.yml migration files are present

hobu:
As suggested in https://github.com/regro/cf-scripts/issues/3704, raise an error when we encounter any `*.yml` files.

Fixes https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/7029


jakirkham:
Thank you both! 🙏

Hopefully this helps people avoid this pitfall in the future

PR #7348, opened 2025-05-03
add migration for pytorch 2.7

h-vetinari:
Just noticed that we never did the piggyback migrator talked about in #7035 for adding windows support more broadly. We can still do this after this one (and/or wait until our mkl has switched to llvm-openmp).

### DIFF
--- a/recipe/migrations/pytorch27.yaml
+++ b/recipe/migrations/pytorch27.yaml
@@ -1,0 +1,10 @@
+__migrator:
+  build_number: 1
+  commit_message: Rebuild for pytorch 2.7
+  kind: version
+  migration_number: 1
+libtorch:
+- '2.7'
+pytorch:
+- '2.7'
+migrator_ts: 1746308782.6617868


hmaarrfk:
we should wait for the windows builds to get uploaded right?

h-vetinari:
Yes... I mean, we could migrate a few feedstocks that don't support windows yet, but some pretty key packages (e.g. torchvision) would be stuck right away.

However, aarch+CUDA also looks like it's failing to upload (no logs visible yet, but already way over average runtime; waiting for the job to finish to see the logs).

In any case, if it's an issue that takes longer to sort out on the anaconda side, it makes no sense to rush imo

h-vetinari:
For anyone following along, this is blocked on https://github.com/conda/infrastructure/issues/1159

jeongseok-meta:
FYI, the windows builds have been uploaded at the 9th attempt: https://github.com/conda-forge/pytorch-cpu-feedstock/pull/383#issuecomment-2919759583

PR #7356, opened 2025-05-05
Add rav1e 0.7 to global pinnings

hmaarrfk:
A new version of rav1e has been released this last year, moving 0.6.6 to 0.7.1

https://github.com/conda-forge/rav1e-feedstock/pull/12

I don't feel like dealing with the fallback from potential ABI compatibility, so I would prefer to tighten the pin and rebuild the (1 ????) feedstock that depends on it, likely https://github.com/conda-forge/libavif-feedstock

However, due to recent binary incompatibility introduced in https://github.com/conda-forge/rav1e-feedstock/issues/13#issuecomment-2852434748 

I'm hoping to skip the step of the migration and move directly to the global pin

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -900,7 +900,7 @@ qt6_main:
 qtkeychain:
   - '0.15'
 rav1e:
-  - '0.6'
+  - '0.7'
 rdma_core:
   - '55'
 re2:
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -899,6 +899,8 @@ qt6_main:
   - 6.9
 qtkeychain:
   - '0.15'
+rav1e:
+  - '0.6'
 rdma_core:
   - '55'
 re2:


hmaarrfk:
I've been merging a few of these myself, but is it OK to do so? or do we want to wait "1 week" for an other core maintainer to chime in?

h-vetinari (approved):
Fine with this, but what's the current situation with the ABI-break you mentioned? Do the currently most recent builds of 0.7.0(?) have the same ABI as the one expected in 0.7.1?

hmaarrfk:
To be clear:

* 0.6.6 build 2 - live with run_exports 'x'
* 0.6.6 build 3 - broke ABI with build 2 due to some internal rust crate stuff - run_exports 'x.x'
* 0.7.1 build 0 - in PR - not ABI compatible with 0.6.6 (either of them) - run_exports 'x.x'

So i'm kinda stuck with 2 bad version 0.6.6 and i kinda just want to rip off the bandaid since only 1 package uses this.

hmaarrfk:
Rollout plan described in https://github.com/conda-forge/rav1e-feedstock/pull/12

PR #7432, opened 2025-05-29
[WIP] Add CUDA 11.8 manual migrator

jakirkham:
We are planning to drop CUDA 11.8 from the global pinnings as discussed in issue: https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/7404

Following the pattern that we have used in the past to readd old CUDA version like with PR: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/1657

Provide a manual migrator that recipe maintainers can use to readd CUDA 11.8 in their feedstocks.

### DIFF
--- a/.yamllint
+++ b/.yamllint
@@ -3,7 +3,7 @@ extends: default
 rules:
   document-start: disable
   line-length:
-    max: 180
+    max: 260
   indentation:
     spaces: consistent
     indent-sequences: whatever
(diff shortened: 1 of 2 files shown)


h-vetinari:
Thanks for doing this. I agree it's nice to provide a migrator for people to copy (even though I suspect it won't see much use 🤷). Did you test it with the state of #7431[^1]? Due to the zipping change (removing `docker_image`), there's potential for things to behave differently once it's been merged.

[^1]: e.g. `conda smithy rerender -e <local_pinning>`

h-vetinari:
Could you test it one last time now that `cuda_compiler` has also changed in #7431?

PR #7453, opened 2025-06-03
feat: add libcap into osx_arm64.txt

wep21:
Checklist

### DIFF
--- a/recipe/migrations/osx_arm64.txt
+++ b/recipe/migrations/osx_arm64.txt
@@ -654,6 +654,7 @@ lib3mf
 libavif
 libblitz
 libboost-mpi
+libcap
 libcint
 libcomcat
 libconeangle


traversaro:
> @wolfv @hmaarrfk add libcap into osx_arm64 migration, but bot PR is not created. How can I move forward?

You can check the migration state in https://conda-forge.org/status/migration/?name=armosxaddition , in particular with the graph visualization you can see which PR blocks the `libcap`:

image

https://github.com/conda-forge/attr-feedstock/pull/10

traversaro:
> @traversaro Thank you for providing useful information.
> 
> > By the way, are you sure that the library can be compiled at all on macOS? The description seems quite Linux-specific, and even on homebrew it is only available on Linux: https://formulae.brew.sh/formula/libcap.
> 
> Ah, I missed availability on macos of homebrew. I've encountered `fatal error: 'sys/capability.h' file not found` when building `realtime_tools`.

Probably OT here, but from what I see the use of 'sys/capability.h' file in https://github.com/ros-controls/realtime_tools/blob/master/realtime_tools/src/realtime_helpers.cpp#L35 is just a small part of all the linuxism in that file. My guess is that for macOS support, we just need to expand all the preprocessor logic that do not compile part of the code on Windows to also include macos as well. Given the great feedback from the ros2_control community, perhaps a good first step could be to open an issue in the realtime_tools repo.

PR #7474, opened 2025-06-12
Simplify python3 zip by removing pypy and defunct python versions

hmaarrfk:
In the great spirit of shedding mental load, I think it would be great to simplify these migrators and our general python pinnings.

This mostly removes a redundant zip key, and a bunch of old python versions (including pypy) to help keep things understandable for new contributors.

There remains `is_python_min` but I think we can remove that in a different PR to entirely remove the python zip.

Checklist

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -193,7 +193,6 @@ zip_keys:
     - cuda_compiler_version     # [win64 and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
   -
     - python
-    - python_impl
     - is_python_min
   -
     - libarrow
@@ -851,16 +850,12 @@ pulseaudio_daemon:
 pybind11_abi:
   - 4
 python:
-  # part of a zip_keys: python, python_impl, is_python_min
+  # part of a zip_keys: python, is_python_min
   - 3.9.* *_cpython
   - 3.10.* *_cpython
   - 3.11.* *_cpython
   - 3.12.* *_cpython
 python_impl:
-  # part of a zip_keys: python, python_impl, is_python_min
-  - cpython
-  - cpython
-  - cpython
   - cpython
 python_min:
   # minimum supported python version per CFEP-25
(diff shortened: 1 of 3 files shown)


h-vetinari:
> There remains `is_python_min` but I think we can remove that in a different PR to entirely remove the python zip.

I'm not in favour of removing this. For recipes that have a noarch python output among others, it's fragile (and hard to get right) to write jinja-logic so that the output only gets built for python_min.

I find it overall preferable to keep the zip (and be able to use `is_python_min` in selectors easily) over the fact that python is part of a zip.

Zooming out a bit, from my POV, zips are painful primarily when they mix conceptually unrelated quantities that we're forced to combine due to infrastructure constraints (like CUDA version with GCC version, with docker images, etc.).

However, all the quantities remaining in the python zip (after the removal of numpy, thankfully) are intrinsically related to the python version, and so I don't consider the existence of the zip to be much of an issue.

hmaarrfk:
> This has kind of questionable benefit IMO? In the sense that it only buys us 2-3 month of simplification, because once we close the 3.13 migration, we'll need to re-add `python_impl` to the zip.
> 
> Not against it per se, just pointing out that the underlying requirement is not going away.

Perhaps is_python_min is too much to remove, but all present supported python_impl are the same. It’s time to reduce the lines of code here to allow mental space for new ideas

h-vetinari:
> It’s time to reduce the lines of code here to allow mental space for new ideas

Haha, sure. In any case, I had misremembered about the 3.13 thing; it's only in the build string, not in the python_impl:
https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/07e16d1185737f9d5051a1b8903bb37e57f3a9b6/recipe/migrations/python313.yaml#L39-L43
https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/07e16d1185737f9d5051a1b8903bb37e57f3a9b6/recipe/migrations/python313t.yaml#L43-L47

h-vetinari (approved):

hmaarrfk:
Thanks all

PR #7476, opened 2025-06-13
Migrate for CUDA 12.9

### LINKED ISSUE #6980: Adding CUDA 12.9 migration

h-vetinari:
Builds on top of #7005 after the problems there were rendered obsolete by dropping CUDA 11.8 (c.f. #7404, #7431)

As a demo, I've opened https://github.com/conda-forge/pytorch-cpu-feedstock/pull/393 ~though this currently needs a smithy PR (https://github.com/conda-forge/conda-smithy/pull/2335) due to an issue with the variant algebra for exactly the case we want to do here: https://github.com/conda-forge/conda-smithy/issues/2331~

Closes #7005
Closes #6980


h-vetinari:
The smithy changes have landed and were released in 3.50.1 (thanks @beckermr!), so this is ready for review!

PTAL especially @conda-forge/cuda.

Perhaps relevant: there's some weird linker errors in https://github.com/conda-forge/pytorch-cpu-feedstock/pull/393 that seem to be due to the CUDA 12.9 toolchain (or some interaction with it).

hmaarrfk (approved):

hmaarrfk (approved):
I beleive you have addressed all the technical blockers correct.

h-vetinari:
Blockers from the infrastructure side should all be resolved, but we haven't got a passing 12.9 build for pytorch yet, and I'd like to understand what's going wrong in the toolchain there (aside from some input or green light from @conda-forge/cuda on this in general).

Feedstocks that want to build with 12.9 can do so of course (and feedback would be welcome!): simply copy the migrator from this PR, add `use_local: true` and then rerender.

hmaarrfk:
Perfect. I guess my approval is on the structure of this PR and once the PyTorch build is ready this can be merged without further input from me

recipe/migrations/cuda129.yaml
```diff
@@ -0,0 +1,44 @@
...
+    To build for these architectures, maintainers will need to add these to list of architectures
+    that their package builds for.
+    
+    ref: https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#new-features
+
+cuda_compiler_version:         # [((linux and (x86_64 or aarch64)) or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+  - 12.9                       # [((linux and (x86_64 or aarch64)) or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+
+c_compiler_version:            # [(linux and (x86_64 or aarch64)) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+  - 13                         # [(linux and (x86_64 or aarch64)) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
```
bdice, line 38: CUDA 12.8 and 12.9 both support GCC 14. I haven't tracked the GCC 14 migration elsewhere on conda-forge enough to know if this should be bumped to 14 or not.

h-vetinari: I know, and I'm planning to make use of this. If https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/7421 gets merged first, I'll update to 14 here. Or if this PR gets merged first, I'll bump the pin in the `cuda129.yaml` file in the other PR.

h-vetinari:
We finally merged https://github.com/conda-forge/pytorch-cpu-feedstock/pull/393, though on windows we had to downgrade to 12.8 because 12.9 was OOM-ing even on the largest possible machine. I'm fine with keeping this specific to pytorch (which is a beast to build anyway), as long as we're reasonably confident that there are no big unresolved issues with 12.9 on windows. It does seem like the toolchain has a problem (or a regression) there though.

recipe/migrations/cuda129.yaml
h-vetinari: On the topic of GCC 14 (which we'll bump to in a few days), it seems that we maybe should stay on GCC 13 for CUDA 12.9 for now. At least on the pytorch side, this combination ran into issues, [namely](https://github.com/conda-forge/pytorch-cpu-feedstock/pull/393#issuecomment-2993932045)

> Looks like GCC 14 might be premature, at least for pytorch (or at least without turning off `-Wincompatible-pointer-types`):
> 
> ```
>  $SRC_DIR/third_party/XNNPACK/src/f16-conv-hwc2chw/f16-conv-hwc2chw-3x3s2p1c3x4-neonfp16arith-2x2.c:53:62: error: passing argument 1 of 'vld1_dup_u16' from incompatible pointer type [-Wincompatible-pointer-types]
>    53 |   const float16x4_t vmax = vreinterpret_f16_u16(vld1_dup_u16(&params->scalar.max));
>       |                                                              ^~~~~~~~~~~~~~~~~~~
>       |                                                              |
>       |                                                              const xnn_float16 * {aka const _Float16 *}
> In file included from $SRC_DIR/third_party/XNNPACK/src/f16-conv-hwc2chw/f16-conv-hwc2chw-3x3s2p1c3x4-neonfp16arith-2x2.c:8:
> $BUILD_PREFIX/lib/gcc/aarch64-conda-linux-gnu/14.3.0/include/arm_neon.h:13130:31: note: expected 'const uint16_t *' {aka 'const short unsigned int *'} but argument is of type 'const xnn_float16 *' {aka 'const _Float16 *'}
> 13130 | vld1_dup_u16 (const uint16_t* __a)
>       |               ~~~~~~~~~~~~~~~~^~~
> ```
> 
> Curious also that this doesn't seem to be an issue on x64, only on aarch64.
>
> At first glance it appears that the type of `params->scalar.max` gets messed up, because casting from `_Float16` to `uint16` sounds very risky, and the `vld1_dup_u16` in [GCC](https://github.com/gcc-mirror/gcc/blob/master/gcc/config/aarch64/arm_neon.h) really is about integers (so I can't see how it'd be a case of picking the wrong overload).

h-vetinari: So #7421 has been merged now. But for now I'm leaving CUDA 12.9 on GCC 13, until the above issue gets fixed or someone tells me that the issue is somehow specific to pytorch.

h-vetinari:
@conda-forge/cuda can someone please comment whether this is good to go from your end. Several feedstocks are waiting to support the new architectures.

I think the remaining open points encountered specifically on the pytorch feedstock (win+12.9 OOMs by works with 12.8; linux compilation errors when using GCC 14) aren't big enough to be blockers for getting this started.

recipe/migrations/cuda129.yaml
bdice: Sounds good to me!

bdice (approved):
This looks fine to me. Maybe @jakirkham or @carterbox can take a quick peek before merging?

recipe/migrations/cuda129.yaml
```diff
@@ -0,0 +1,47 @@
...
+    Upgrade to CUDA 12.9
+    
+    CUDA 12.8 added support for architectures `sm_100`, `sm_101` and `sm_120`,
+    while CUDA 12.9 further added `sm_103` and `sm_121`. To build for these,
+    maintainers will need to modify the list of specified architectures
+    (e.g. `CMAKE_CUDA_ARCHITECTURES`, `TORCH_CUDA_ARCH_LIST`, etc.)
+    for their package. A good balance between broad support and storage
+    footprint (resp. compilation time) is to add `sm_100` and `sm_120`.
+    
+    ref: https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#new-features
```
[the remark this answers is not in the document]
bdice: Ah! That is much clearer. Thanks @carterbox.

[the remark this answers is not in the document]
h-vetinari: Thanks!

h-vetinari:
Alright, thanks for the inputs @bdice @carterbox. I'll merge this in 72h unless there are other comments.

carterbox (approved):

recipe/migrations/cuda129.yaml
h-vetinari: > [...] or someone tells me that the issue is somehow specific to pytorch.

FYI, since it turns out that the problems were specific to pytorch, I'm bumping the CUDA 12.9 to GCC 14 now, to match the rest of the pinning: #7563

PR #7482, opened 2025-06-13
fix cuda_compiler_version_min in CUDA 11.8 migrator

h-vetinari:
Fix-up for #7472 while applying this to arrow; also needs https://github.com/conda-forge/conda-smithy/pull/2338 to work correctly.


h-vetinari:
OK, this is a trivial fix, and tested in https://github.com/conda-forge/arrow-cpp-feedstock/pull/1773, so in it goes.

PR #7565, opened 2025-07-15
delete ruby from pinning

h-vetinari:
The ruby 3.1 migration has been open for 3 years, and nothing is happening (xref #4805). This is a continuous waste of compute cycles in the bot infrastructure for no gain. Furthermore, the existing pins are completely out of date (i.e. still require openssl<3, which we've been on for years at this point).

So just remove ruby from the pinning and be done with it. Small focussed efforts (e.g. on `ruby` itself and a few core packages) can still continue of course. But for it to be part of the pinning, someone would have to bring sustained effort to the table.

CC @conda-forge/ruby @jdmarques


danielnachun:
Once we've added support for recipe generation and bot updates from rubygems.org, I'd like to start a new migration for the 3 supported versions - 3.2, 3.3, and 3.4.  But it makes no sense to keep trying to migrate to a now EOL version of Ruby.

h-vetinari:
@danielnachun that's a good plan, if you're up for it! :)

In the meantime, you could approve this PR ;-)

danielnachun (approved):

hmaarrfk:
is it ok to leave the pin floating? these kinds of floating pins are hard to back out of, and would effectively bump all the way to 3.4

h-vetinari:
The ruby ecosystem (outside of `ruby` itself) in conda-forge is dead as a doornail. So already on that level, I consider it irrelevant whether we float or not.

> these kinds of floating pins are hard to back out of, and would effectively bump all the way to 3.4

Aside from the above, and aside from finally becoming up-to-date from so-EOL-they're-unusable versions (so what?), it's also quite possible (for the sake of argument) to migrate for 3.2 after some packages have been built for 3.4 already.

So an effort to reintroduce the pin would either have to use the latest version, or do a full migration to get a consistent lower pin. Neither is a problem.

h-vetinari:
> What if we pinned to 3.2 without any migration for now?

I don't have a strong opinion, but I'd prefer to do this with a proper migration, and then either update ruby-dependent feedstocks for that, or archive those. Only switching the pin to 3.2 is IMO just another bandaid on a sucking chest wound.

> Also this highlights the problem that we are shipping a very EOL version of Perl

Yup, another quasi-abandoned ecosystem. I'm glad you're so motivated to keep all major (and even not-so-major) languages in shape!

danielnachun:
> Yup, another quasi-abandoned ecosystem. I'm glad you're so motivated to keep all major (and even not-so-major) languages in shape!

Ironically I have found that Perl, Ruby and Lua are **significantly** easier ecosystems to maintain with Conda packages than Python.  So I very much want to see them get to the same quality of support as Python as I think maintaining feedstocks for those ecosystems is substantially more automation-friendly than Python.

h-vetinari:
> I think maintaining feedstocks for those ecosystems is substantially more automation-friendly than Python.

That would be great! Much of the R ecosystem is also automated (your work there notwithstanding, and the effort for stuff like the UCRT migration), so there's precedent for this to work without much supervision.

PR #7619, opened 2025-07-26
Release harfbuzz pin

hmaarrfk:
I feel like the old pin 11.0.1 was quite restrictive and harfbuzz has had many releases since.

Most notably 11.3 which adds some new API/ABI

Just wanting this to keep up.

I think we may even be able to remove harfbuzz from the global pins alltogether tbh.

LMK

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -370,11 +370,8 @@ googleapis_cpp:
   - '0.10'
 graphviz:
   - '13'
-# Harfbuzz guarantees total ABI compatibiblity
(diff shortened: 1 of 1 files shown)


h-vetinari (approved):
From the linked issue, a link to upstream [saying](https://github.com/harfbuzz/harfbuzz/discussions/5245#discussioncomment-12728609):
> We take the API and ABI stability seriously. 0.9.30 was released in 2014, so we haven’t changed the API or ABI incompatibly in 11 years.

hmaarrfk:
Il going to close this PR to open a new one where I just delete harfbuzz from here altogether.

PR #7653, opened 2025-08-12
NEW: Migrate recipes to CUDA 13.0

carterbox:
Checklist

## Test Feedstocks:

- [ ] https://github.com/conda-forge/libastra-feedstock/pull/35 Rough edge with Windows binary paths
- [ ] ~~NVCOMP~~ Not built from source
- [ ] https://github.com/conda-forge/ucx-split-feedstock/pull/222 (unix only!)
- [ ] https://github.com/conda-forge/cupy-feedstock/pull/314 Rough edge with Windows binary paths

## Blocking Issues:

- [x] https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/397
- [x] https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/395

### DIFF
--- a/recipe/migrations/cuda118.yaml
+++ b/recipe/migrations/cuda118.yaml
@@ -46,6 +46,9 @@ cuda_compiler_version:      # [(linux or win64) and os.environ.get("CF_CUDA_ENAB
 cuda_compiler_version_min:  # [linux or win64]
   - 11.8                    # [linux or win64]
 
+c_stdlib_version:           # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+  - 2.17                    # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+
 c_compiler_version:         # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
   - 11                      # [linux and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
 
--- a/recipe/migrations/cuda118.yaml
+++ b/recipe/migrations/cuda118.yaml
@@ -28,6 +28,7 @@ __migrator:
       - 12.8
       - None
       - 12.9
+      - 13.0
       - 11.8
     cuda_compiler_version_min:
       - 12.4
--- a/recipe/migrations/cuda118.yaml
+++ b/recipe/migrations/cuda118.yaml
@@ -1,11 +1,11 @@
-migrator_ts: 1748496951
+migrator_ts: 1755549433
 __migrator:
   kind:
     version
   migration_number:
     1
   build_number:
-    1
+    2
   # This is intended as a _manual_ migrator to re-add CUDA 11.8, after we
   # dropped it as version that's built by default; DO NOT unpause
   paused: true
--- a/recipe/migrations/cuda118.yaml
+++ b/recipe/migrations/cuda118.yaml
@@ -1,4 +1,5 @@
-migrator_ts: 1755549433
+# In order to work, this timestamp needs to be newer than all other CUDA migrators
+migrator_ts: 2145938400  # 2038-01-01
 __migrator:
   kind:
     version
--- a/recipe/migrations/cuda118.yaml
+++ b/recipe/migrations/cuda118.yaml
@@ -6,7 +6,7 @@ __migrator:
   migration_number:
     1
   build_number:
-    2
+    1
   # This is intended as a _manual_ migrator to re-add CUDA 11.8, after we
   # dropped it as version that's built by default; DO NOT unpause
   paused: true
--- a/recipe/migrations/cuda130.yaml
+++ b/recipe/migrations/cuda130.yaml
@@ -4,7 +4,7 @@ __migrator:
   migration_number:
     1
   build_number:
-    0
+    1
   paused: false
   override_cbc_keys:
     - cuda_compiler_stub
--- a/recipe/migrations/cuda130.yaml
+++ b/recipe/migrations/cuda130.yaml
@@ -26,7 +26,7 @@ __migrator:
     - cuda129
   commit_message: |
     Upgrade to CUDA 13.0
-
+    
     CUDA 13.0 requires architecture `sm_75` or higher, and renamed `sm_101` to
     `sm_110`. To build for these, maintainers will need to modify their existing list of
     specified architectures (e.g. `CMAKE_CUDA_ARCHITECTURES`, `TORCH_CUDA_ARCH_LIST`, etc.)
--- a/recipe/migrations/cuda130.yaml
+++ b/recipe/migrations/cuda130.yaml
@@ -44,9 +44,6 @@ __migrator:
 cuda_compiler_version:         # [((linux and (x86_64 or aarch64)) or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
   - 13.0                       # [((linux and (x86_64 or aarch64)) or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
 
-cuda_compiler_version_min:     # [((linux and (x86_64 or aarch64)) or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
-  - 12.9                       # [((linux and (x86_64 or aarch64)) or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
-
 c_stdlib_version:              # [(linux and (x86_64 or aarch64)) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
   - 2.28                       # [(linux and (x86_64 or aarch64)) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
 
--- a/recipe/migrations/cuda130.yaml
+++ b/recipe/migrations/cuda130.yaml
@@ -27,7 +27,7 @@ __migrator:
   commit_message: |
     Upgrade to CUDA 13.0
 
-    CUDA 13.0 removed support for architectures before `sm_75`, and renamed `sm_101` to
+    CUDA 13.0 requires architecture `sm_75` or higher, and renamed `sm_101` to
     `sm_110`. To build for these, maintainers will need to modify their existing list of
     specified architectures (e.g. `CMAKE_CUDA_ARCHITECTURES`, `TORCH_CUDA_ARCH_LIST`, etc.)
     for their package.
--- a/recipe/migrations/cuda130.yaml
+++ b/recipe/migrations/cuda130.yaml
@@ -1,8 +1,6 @@
 migrator_ts: 1755016036
 __migrator:
-  use_local: true
-  kind:
-    version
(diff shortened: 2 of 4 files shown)


recipe/migrations/cuda130.yaml
```diff
@@ -0,0 +1,64 @@
...
+  - None                       # [((linux and (x86_64 or aarch64)) or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+  - 12.9                       # [((linux and (x86_64 or aarch64)) or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+  - 13.0                       # [((linux and (x86_64 or aarch64)) or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+
+cuda_compiler_version_min:     # [((linux and (x86_64 or aarch64)) or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+  - 12.9                       # [((linux and (x86_64 or aarch64)) or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+
+c_compiler_version:            # [(linux and (x86_64 or aarch64)) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+  - 14                         # [(linux and (x86_64 or aarch64)) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+  - 15                         # [(linux and (x86_64 or aarch64)) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
```
h-vetinari, line 56: We're not using GCC 15 anywhere yet. It's great that CUDA 13.0 can handle it, but this will only become relevant next ~June

h-vetinari (commented):
looks pretty good already!

recipe/migrations/cuda130.yaml
```diff
@@ -0,0 +1,64 @@
...
+      - 12.8
+      - None
+      - 12.9
+      - 13.0
+      # to allow manual opt-in for CUDA 11.8, see
+      # https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/7472
+      # must be last due to how cuda_compiler ordering in that migrator works
+      - 11.8
+  wait_for_migrators:
+    - cuda129
```
h-vetinari, line 28: On second thought, this should very likely use `operation: key_add`. cuda129.ymal updates in place, but cuda13.yaml extends the zip.

recipe/migrations/cuda130.yaml
```diff
@@ -0,0 +1,64 @@
...
+    version
+  migration_number:
+    1
+  build_number:
+    0
+  paused: false
+  override_cbc_keys:
+    - cuda_compiler_stub
+  check_solvable: false
+  primary_key: cuda_compiler_version
```
h-vetinari, line 14: Speaking of the zip, we will need to modify the global cuda zip to include `c_stdlib_version`, and then override that value in the migrator.

recipe/migrations/cuda130.yaml
```diff
@@ -0,0 +1,64 @@
+migrator_ts: 1755016036
+__migrator:
+  use_local: true
```
h-vetinari, line 3: This should not be in the global migrator

h-vetinari (commented):
Since we haven't closed the 12.9 migration yet, it would be good to test that the presence of _both_ `cuda129.yaml` and `cuda130.yaml` does the right thing.

recipe/migrations/cuda130.yaml
carterbox: The documentation isn't great, but I though `key_add` was for adding new keys e.g. `c_compiler_version` not versions themselves e.g. `15`. Let me try again to get a separate migrator for CUDA 13.0. Otherwise, the current approach of this PR is to immediately close the CUDA 12.9 migrator and open this one which does both 12.9 and 13.0 simultaneously.

recipe/migrations/cuda130.yaml
carterbox: Removed in 5b0c52a3370e75258a60028c95658475f57f4bcf

recipe/migrations/cuda130.yaml
carterbox: `key_add` seems to be working how we want. You can apply the 130 migrator on top of the 129 migrator and it keeps both.

5b0c52a3370e75258a60028c95658475f57f4bcf

recipe/migrations/cuda130.yaml
carterbox: Removed in 5b0c52a3370e75258a60028c95658475f57f4bcf

h-vetinari:
With the changes to the global zip this becomes a bit harder to test locally, but it's still possible. You need to use the `recipe/conda_build_config.yaml` from this PR, copy it somewhere (say that file is called `cbc`), then you can test rerender this migrator (with `use_local: true`, and the relevant changes to cuda129.yaml if present) using `conda smithy rerender -e ../cbc`

At first I wanted to use `additional_zip_keys:` like for the nogil migrator
https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/c456e01b56a7406268e582d0578097c90578a207/recipe/migrations/python313t.yaml#L27-L29
This would postpone having to modify the global pinning until we close the 13.0 migration. But I wasn't able to get this work, it fails in smithy with

```
  File "E:\miniforge\envs\builder\Lib\site-packages\conda_smithy\variant_algebra.py", line 191, in op_variant_key_add
    additional_zip_keys_default_values[additional_key] = result[additional_key][0]
                                                         ~~~~~~^^^^^^^^^^^^^^^^
KeyError: 'c_stdlib_version'
```

Since the question is just _when_ we add `c_stdlib_version` back to the zip, I just went ahead and did it here. I was able to rerender the pytorch feedstock with these changes, and it did the correct thing.

recipe/migrations/cuda130.yaml
```diff
@@ -0,0 +1,57 @@
...
+      # to allow manual opt-in for CUDA 11.8, see
+      # https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/7472
+      # must be last due to how cuda_compiler ordering in that migrator works
+      - 11.8
+  wait_for_migrators:
+    - cuda129
+  commit_message: |
+    Upgrade to CUDA 13.0
+
+    CUDA 13.0 removed support for architectures before `sm_75`, and renamed `sm_101` to
```
h-vetinari, line 30: Would be good to be explicit whether `sm_75` is dropped or still supported. Otherwise, people will likely interpret "before" inconsistently, as either `<`  or `≤`.

h-vetinari: e.g.
```suggestion
    CUDA 13.0 removed support for architectures `sm_72` and below, and renamed `sm_101` to
```

msarahan: I think sm_75 is Turing, and is still supported. I think the edit here makes good sense.

carterbox: Chose Bradley's wording in 521045c8abe8588cf3ff7952f9946a2d79a16c2d

h-vetinari (approved):
From my POV, this is good to go

bdice (approved):
Seems fine to me!

carterbox:
There's still some bugs in the Windows CUDA 13.0 build for libastra which AFAIK is the only compiled feedstock that this migrator has been tested against. Until we have a few examples of non-redist packages successfully building with the CUDA 13.0 CTK, I'd prefer not to start the migration.

recipe/migrations/cuda130.yaml
```diff
@@ -0,0 +1,62 @@
...
+      - 13.0
+      # to allow manual opt-in for CUDA 11.8, see
+      # https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/7472
+      # must be last due to how cuda_compiler ordering in that migrator works
+      - 11.8
+  wait_for_migrators:
+    - cuda129
+  commit_message: |
+    Upgrade to CUDA 13.0
+
```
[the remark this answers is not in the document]
carterbox: Wow that's a rough edge. Has anyone ever tried adding the indentation indicator?

e.g.

```yaml
commit_message: |2
  Some text here
```

recipe/migrations/cuda130.yaml
```diff
@@ -0,0 +1,62 @@
+migrator_ts: 1755016036
+__migrator:
+  operation: key_add
+  migration_number:
+    1
+  build_number:
+    0
+  paused: false
```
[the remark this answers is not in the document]
bdice: Agreed. We could pause it at first as we iron out some kinks and then unpause it.

recipe/migrations/cuda130.yaml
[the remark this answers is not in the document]
h-vetinari: I don't remember the details, but yes, this is relatively fragile. Luckily very few migrators have a multi-line commit message. Implementation can be found in this [module](https://github.com/regro/cf-scripts/blob/c96fa978095baac082b8b227ef12da208e9ba4cb/conda_forge_tick/migrators/migration_yaml.py#L539) (and elsewhere in that repo, though I believe this is the most relevant module).

recipe/migrations/cuda118.yaml
```diff
@@ -1,11 +1,11 @@
-migrator_ts: 1748496951
+migrator_ts: 1755549433
```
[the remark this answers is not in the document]
h-vetinari: Fine by me!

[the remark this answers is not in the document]
h-vetinari: Please also add an explanatory comment behind whatever date you choose, e.g.
```suggestion
migrator_ts: xxxxxxxxxx  # 2023-01-01
```

recipe/migrations/cuda130.yaml
```diff
@@ -0,0 +1,62 @@
...
+  wait_for_migrators:
+    - cuda129
+  commit_message: |
+    Upgrade to CUDA 13.0
+    
+    CUDA 13.0 requires architecture `sm_75` or higher, and renamed `sm_101` to
+    `sm_110`. To build for these, maintainers will need to modify their existing list of
+    specified architectures (e.g. `CMAKE_CUDA_ARCHITECTURES`, `TORCH_CUDA_ARCH_LIST`, etc.)
+    for their package.
+
```
carterbox, line 34: Insert note here about binary path changes on Windows

h-vetinari: Do you want to do this before merging the still-paused migration? We could also add that text later.

h-vetinari (approved):
LGTM, thanks!

recipe/migrations/cuda130.yaml
carterbox: No, I don't want to add the note because we may be reverting the path changes on Windows.

carterbox:

h-vetinari:
As discussed in the core call. Let's merge as paused, and figure out the remaining issues before unpausing.

PR #7661, opened 2025-08-13
close python 3.13 migration; drop python 3.9

### LINKED ISSUE #7597: Close Python 3.13 migrator with final release of Python 3.14
This sounds like a good time to close the Python 3.13 migration. @conda-forge/core Can someone please also raise this in the core call? I won't be able to join in the next weeks.

Note that this issue is independent of the `python313t` migration, which should be discussed in a separate issue.

h-vetinari:
We've adopted the pattern of dropping `v{N-5}` when we start migrating for the rc's of `vN` with 3.12 & 3.13 already, and have documented this in https://github.com/conda-forge/python-feedstock/issues/794.

This prepares the path for merging (one of) #7598 or #7648; this PR should be done just prior to merging that.

Closes #7597


hmaarrfk:
> We've adopted the pattern of dropping v{N-5} 

Would it help to have a CFEP for this?

h-vetinari:
FWIW, my "👍" is intended as: "sure we can can have a CFEP about this." I don't think we should block this PR on that though.

chrisburr (approved):

h-vetinari:
Announcement was merged, and no opposition in the core call

PR #7687, opened 2025-08-22
DO NOT MERGE: unpause 3.14 migration

h-vetinari:
I'm just opening this for visibility and for being able to track the issues we need to solve before we restart the migration again (i.e. undo #7686 to continue #7598).

* [ ] https://github.com/conda-forge/cross-python-feedstock/issues/98
* [ ] https://github.com/conda-forge/conda-smithy/issues/2366
* [x] #7686 (fixes this [issue](https://github.com/conda-forge/conda-smithy/issues/2367))

### DIFF
--- a/recipe/migrations/python314.yaml
+++ b/recipe/migrations/python314.yaml
@@ -18,7 +18,7 @@ __migrator:
             - 3.13.* *_cp313
             - 3.13.* *_cp313t
             - 3.14.* *_cp314  # new entry
-    paused: true
+    paused: false
     longterm: true
     pr_limit: 5
     max_solver_attempts: 3  # this will make the bot retry "not solvable" stuff 12 times


beckermr (changes requested):
Requesting changes so we have a discussion before merging.

recipe/migrations/python314.yaml
```diff
@@ -6,7 +6,7 @@
 migrator_ts: 1724712514
```
h-vetinari, line 6: In https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/7598 this was initially merged as
https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/861cb54c67058d9b5bf7e91f644e0dfc30958f86/recipe/migrations/python314.yaml#L4

which notably collides with the timestamp that was used for `python313.yaml` before it was deleted
https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/565bd4c5914b25e189c5c57888e66a86b52d83e5/recipe/migrations/python313.yaml#L1

Then in 4a465c585afabf700324db086ba49b20ef910e81 we went to
https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/031eab9e3ca8f6cbfd402ca235bf4393f54962a3/recipe/migrations/python314.yaml#L6

which was then changed again in 2eba9c94a836c15b5a562bdf97bdcad7477a19d7 to
https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/02d4dc6eb31b5e5e5b1539cb2255c6a9a88936e4/recipe/migrations/python314.yaml#L6

However, now that https://github.com/conda-forge/conda-smithy/pull/2374 switched away from mixing up migrations if the timestamps match, I think we can actually go back to the timestamp which we used initially.

recipe/migrations/python314.yaml
```diff
@@ -6,7 +6,7 @@
...
 __migrator:
     commit_message: Rebuild for python 3.14
-    migration_number: 1
+    migration_number: 2
```
h-vetinari, line 9: That's because I dislike bumping `migration_number` - not even due to the extra PR churn, but because it becomes much more confusing for the average user why a given migration (e.g. with `migration_number: 1`) suddenly doesn't apply anymore. I think we should try to avoid this for such a big and long-running migration as 3.14 will be.

beckermr:
No sorry. We discussed on Zulip but did not make an issue.

h-vetinari:
Can you link the zulip discussion? I'm only aware of the issues mentioned in https://github.com/conda-forge/conda-smithy/issues/2366, which are fixed with the new smithy release.

beckermr:
This channel zip error is happening during the solver checks in the bot. I wonder if that code path is hitting a special case or something.

h-vetinari:
> This channel zip error is happening during the solver checks in the bot

Worth noting also that numpy (the feedstock where I've seen that error in the status page) rerenders fine with the current 3.14 migrator.

beckermr:
OK let's go ahead and see what happens with this migration turned back on.

h-vetinari:
Let's squash

PR #7691, opened 2025-08-22
Add check for migrator filename uniqueness

h-vetinari:
This is WIP; see https://github.com/conda-forge/conda-smithy/pull/2369

### DIFF
--- a/.github/workflows/migrator_uniqueness.yml
+++ b/.github/workflows/migrator_uniqueness.yml
@@ -1,0 +1,51 @@
+name: Check Migration Filename Uniqueness
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  migration-check:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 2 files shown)


h-vetinari:
> See my comment on the smithy issue. GH search only covers up to 4000 feedstocks. So this approach won't work.

what you quoted is that it will _find_ only up to 4000 repos. But we only need the search to return a single hit for things to fail. The limit is not relevant for this detection.

beckermr:
I am almost 100% sure the text implies github won't search all of our repositories. It says

> To keep the REST API fast for everyone, we limit the number of repositories a query will search through.

They are not limiting the results, but instead the number of repos searched.

h-vetinari:
> They are not limiting the results, but instead the number of repos searched.

You left out the following sentence which makes it clear this is about search results. "The REST API will find up to 4,000 repositories that match your filters and return results from those repositories."

Their API even provides a response whether the search results are complete
```
Got {
  "total_count": 0,
  "incomplete_results": false,      # <----
  "items": [
  ]
}
```

I strongly doubt that they can even technically limit the number of repos they search, because it's all [based](https://github.blog/engineering/architecture-optimization/the-technology-behind-githubs-new-code-search/) on a giant index, and not individually walking the repos.

PR #7725, opened 2025-09-01
Update pin for impi_devel to 2021.16.0

minrk:
use x.y.z to match [run_exports](https://github.com/conda-forge/intel_repack-feedstock/blob/ae56ebee50a9e4e9e0c166b62804e444ca7243a9/recipe/meta.yaml#L947) for impi_devel, since past patch releases without migrators have caused problems (e.g. 2021.14.1).

closes #7271 which seems to have been made in error, since there is no 2025 release of impi_devel.

### DIFF
--- a/recipe/migrations/impi202116.yaml
+++ b/recipe/migrations/impi202116.yaml
@@ -1,0 +1,8 @@
+migrator_ts: 1756697132
+__migrator:
+  kind: version
+  migration_number: 1
+  bump_number: 1
+
+impi_devel:
+  - "2021.16.0"


minrk:
Sorry, forgot to copy the justification from #7271: This matches impi_devel's [run_exports](https://github.com/conda-forge/intel_repack-feedstock/blob/28b063f8cde0c4ab139ac56d8eb0763ac44bd021/recipe/meta.yaml#L947). If the pinning is looser than run_exports, we will have 2021.16.0 builds and 2021.16.1 builds, which are not mutually _installable_, and no migrator to fix that. I can't speak to whether the package's run_exports should be looser (perhaps they should), but the pinning here does need to be at least as tight on the upper bound to avoid the problems we have had with 2021.14.1.

xaleryb:
Hi @minrk, 
Thanks for the clarifications. Everything you explained makes sense, and I don’t have any concerns. I just want to add that the situations you described only occur when different components use different levels of pinning — for example, if one component uses x.x while another in the same dependency chain or stack uses x.x.x. That’s where conflicts arise. I believe we’re on the same page here.

In my previous message, I also wanted to highlight that hard pinning isn’t only about control — it brings additional maintenance overhead as well, even if it’s technically safer.

Before we merge the PR: I noticed your PR uses impi_devel, but the conda-forge channel has it as impi-devel (underscore vs. dash).

PR #7757, opened 2025-09-09
include noarch in R migration

mfansler:
We usually run as `longterm` and need to migrate `noarch` packages. I'm satisfied with all the recipes and runs I've inspected, so I'm good for enabling automerge.

CC: @conda-forge/r

For quick ref, this was the R 4.4 migration:

https://github.com/mfansler/conda-forge-pinning-feedstock/blob/f403d3805e818106bc556fb793916088d58d7683/recipe/migrations/r-base44_and_m2w64-ucrt.yaml

### DIFF
--- a/recipe/migrations/r_base45.yaml
+++ b/recipe/migrations/r_base45.yaml
@@ -4,6 +4,10 @@ __migrator:
   kind: version
   migration_number: 1
   pr_limit: 5
+  primary_key: r_base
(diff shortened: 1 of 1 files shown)


mfansler:
Note if we don't turn on `noarch`, we'll start accumulating failures. E.g., https://github.com/conda-forge/r-diffobj-feedstock/pull/18

PR #7795, opened 2025-09-19
restart python 3.14 migration due bytecode invalidation in rc3

h-vetinari:
I dislike bumping `migration_number:`s, because it makes the mechanics of "when/how does this migrator apply" much less obvious for maintainers.

However, in this case I think our hands are tied. Python has compiled bytecode (`.pyc` files) which contain a magic number that determines whether they're consumable for a given runtime. In 3.14.0rc3 this magic number was [bumped](https://discuss.python.org/t/python-3-14-0rc3-is-go/103815) _again_ (after we had already waited for rc2 because that also bumped it).

This doesn't break the existing artefacts, but it invalidates their bytecode, meaning any user of py314 that downloads one of the packages with the old magic number will have to wait while a bunch of code gets recompiled, which will be quite visible as a regression of start-up time in a new environment.

Since we're expressly publishing builds intended to be compatible with 3.14 GA already, the only way to ensure that we have a clean set of packages is to rebuild everything. This is painful now, but will only get more painful in the future, so let's bite the bullet sooner rather than later.

Draft until https://github.com/conda-forge/python-feedstock/pull/814 is in.

CC @conda-forge/core

### DIFF
--- a/recipe/migrations/python314.yaml
+++ b/recipe/migrations/python314.yaml
@@ -4,7 +4,7 @@
 migrator_ts: 1724712607
 __migrator:
     commit_message: Rebuild for python 3.14
-    migration_number: 1
+    migration_number: 2
     operation: key_add
     primary_key: python
     ordering:


hmaarrfk:
I don’t really see this regression as bad enough to restart the migration. 

Noarch python also suffer from this regression. 

It sucks, but bytecode compilation right?

h-vetinari:
As soon as https://github.com/conda-forge/python-feedstock/pull/814 is merged (or if we add artefact persistence to that PR), we could do some timing tests of python 3.14.0 rc2 vs rc3, in an environment full of builds against the old magic number. 

It's always helpful to have some concrete metrics for decision-making. Perhaps I'm way off, but my intuition is that for larger environments (of packages that are all being actually imported in the code under test), we're talking about potentially several minutes or longer being added to the first start up. 

If we're in that ballpark, I don't think that'd be an acceptable regression TBH, made worse by the fact that a lot of rarely-rebuilt feedstocks will keep that regression for a long time. 

PS. I commented in the discourse thread I linked in the OP, discussion is ongoing (basically: this shouldn't happen so late, but not doing it was considered worse).

h-vetinari:
I did
```
mamba create -n timing -c conda-forge -c conda-forge/label/python_rc python=3.14.0rc2 pillow scipy matplotlib numpy
```
created a little script to exercise the loading of bytecode by importing stuff

<details>
<summary><code>imports.py</code></summary>

```python
# optimistic star imports
from numpy import *
from scipy import *
from matplotlib import *
from PIL import *
# try to force exhaustive imports
import numpy.dtypes
import numpy.exceptions
import numpy.f2py
import numpy.fft
import numpy.lib
import numpy.lib.format
import numpy.lib.mixins
import numpy.lib.recfunctions
import numpy.lib.scimath
import numpy.lib.stride_tricks
import numpy.lib.npyio
import numpy.lib.introspect
import numpy.lib.array_utils
import numpy.linalg
import numpy.ma
import numpy.ma.extras
import numpy.ma.mrecords
import numpy.polynomial
import numpy.polynomial.chebyshev
import numpy.polynomial.hermite
import numpy.polynomial.hermite_e
import numpy.polynomial.laguerre
import numpy.polynomial.legendre
import numpy.polynomial.polynomial
import numpy.random
import scipy.cluster
import scipy.cluster.hierarchy
import scipy.cluster.vq
import scipy.constants
import scipy.datasets
import scipy.fft
import scipy.fftpack
import scipy.integrate
import scipy.interpolate
import scipy.io
import scipy.io.arff
import scipy.io.matlab
import scipy.io.wavfile
import scipy.linalg
import scipy.linalg.blas
import scipy.linalg.cython_blas
import scipy.linalg.cython_lapack
import scipy.linalg.interpolative
import scipy.linalg.lapack
import scipy.misc
import scipy.ndimage
import scipy.odr
import scipy.optimize
import scipy.signal
import scipy.signal.windows
import scipy.sparse
import scipy.sparse.csgraph
import scipy.sparse.linalg
import scipy.spatial
import scipy.spatial.distance
import scipy.spatial.transform
import scipy.special
import scipy.stats
import scipy.stats.contingency
import scipy.stats.distributions
import scipy.stats.mstats
import scipy.stats.qmc
import scipy.stats.sampling
import PIL.Image
import PIL.ImageCms
import matplotlib.backends._backend_agg
import matplotlib.ft2font
import matplotlib.pyplot
import mpl_toolkits
import pylab
```

</details>
measured the timing with rc2, then updated the same environment to rc3, and measured again. On linux I got
```
$ time python imports.py

# rc2
real    0m2.185s
user    0m2.971s
sys     0m0.259s

# rc3
real    0m4.389s
user    0m5.306s
sys     0m0.230s
```
and on windows
```
Measure-Command { python imports.py }
[...]
TotalSeconds      : 1.7986322  # before
TotalSeconds      : 9.2766205  # after
```

So it's certainly measurable, but less than I had feared. On the other hand, I only had a pretty small-ish environment (many big packages aren't available for 3.14 yet), so in an environment that imports 10s or 100s of packages, this will probably much more noticeable.

PR #7825, opened 2025-10-03
DOC: Remove mention of torch list from migration

carterbox:
The nvcc packages will no longer set TORCH_CUDA_ARCH_LIST in an activation script because pytorch explicitly checks whether the provided archs are valid. This causes the list from new compilers to be incompatible with older versions of pytorch when the list contains very new archs.

https://github.com/conda-forge/cuda-nvcc-feedstock/pull/80
https://github.com/conda-forge/cuda-nvcc-feedstock/pull/79

Checklist

### DIFF
--- a/recipe/migrations/cuda129.yaml
+++ b/recipe/migrations/cuda129.yaml
@@ -3,7 +3,7 @@ __migrator:
   kind:
     version
   migration_number:
-    1
+    2
   build_number:
     1
   paused: false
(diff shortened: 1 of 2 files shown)


bdice (approved):

recipe/migrations/cuda129.yaml
```diff
@@ -3,7 +3,7 @@ __migrator:
   kind:
     version
   migration_number:
-    1
```
carterbox, line 6: Looking for guidance on whether this migration_number should be incremented or not.

recipe/migrations/cuda129.yaml
```diff
@@ -3,7 +3,7 @@ __migrator:
...
+    2
```
carterbox, line 6:
```suggestion
    1
```

recipe/migrations/cuda130.yaml
```diff
@@ -2,7 +2,7 @@ migrator_ts: 1755016036
 __migrator:
   operation: key_add
   migration_number:
-    1
+    2
```
carterbox, line 5:
```suggestion
    1
```

recipe/migrations/cuda129.yaml
beckermr: Ok then don't bump the migration number.

h-vetinari (approved):

PR #7930, opened 2025-11-05
Revert "add migration for cudnn 9.10"

h-vetinari:
Rather than close #7917 as usual, we'll be able to revert #7891, and go back to the regular cudnn pinning, as soon as
https://github.com/conda-forge/admin-requests/pull/1742 has landed without causing any major breakage.

The feedstock-local changes (of adding `cudnn:`) in overridden migrators can remain without causing issues, because the cudnn 9.10 pin was kept out of those changes, and confined to the migrator being deleted here.

Closes #7917


h-vetinari:
Let's give https://github.com/conda-forge/admin-requests/pull/1742 at least 48h to see if there's yet more fallout we need to address before going through with this.

PR #7936, opened 2025-11-06
addr-fst

ZerryNi:


ZerryNi:
> Please don't create new PRs for the same thing. You should push new commits (or force-push them, if you prefer) into #7923, not create new PRs for a minor change. This makes it very hard to have continuity in the review process.

okay,I'am sorry. but I wonder if xorg-libxcb really cant't be added?

PR #7999, opened 2025-11-27
also pin libcudnn-dev

h-vetinari:
cudnn has a more granular set of outputs now that matches the rest of the CUDA `lib*` packages. This also causes the link check to complain, because now the library is actually in `libcudnn`, which gets run-exported from `libcudnn-dev`, whereas `cudnn` simply run-exports `cudnn`.

I'm switching (resp. planning to) a few feedstocks from `cudnn` to `libcudnn-dev`, and of course the pinning should remain when doing so.

CC @conda-forge/cudnn

### DIFF
--- a/recipe/migration_support/packages_to_migrate_together.yaml
+++ b/recipe/migration_support/packages_to_migrate_together.yaml
@@ -30,6 +30,10 @@ libboost_devel:
   - libboost_headers
   - libboost_python_devel
 
+libcudnn_dev:
+  - libcudnn_dev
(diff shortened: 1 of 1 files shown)


hmaarrfk:
should we add this to https://github.com/h-vetinari/conda-forge-pinning-feedstock/blob/cudnn/recipe/migration_support/packages_to_migrate_together.yaml

i know we likely won't do "auto" migrations, but maybe it helps? or maybe its just noise....

PR #8004, opened 2025-11-30
Add migration for libflac 1.5

hmaarrfk:
it was never added to the global migrators, so packages were never rebuilt in the last year

libsndfile still depends on 1.4


hmaarrfk:
cc: @conda-forge/libflac

PR #8029, opened 2025-12-05
Migrate for aws-c-* dec'25-3

xhochy:
aws-c-* migration for dec'25-3

## Updated packages:
- aws_c_auth: 0.9.3
- aws_c_event_stream: 0.5.7
- aws_c_s3: 0.11.3
- s2n: 1.6.2

### DIFF
--- a/recipe/migrations/aws_c_dec253.yaml
+++ b/recipe/migrations/aws_c_dec253.yaml
@@ -1,0 +1,16 @@
+__migrator:
+  build_number: 1
+  commit_message: Rebuild for aws-c-* (dec'25-3)
+  kind: version
+  migration_number: 1
+  exclude_pinned_pkgs: false
+  automerge: true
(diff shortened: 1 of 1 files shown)


h-vetinari:
The determining factor is which version of the pinning is pinned in https://github.com/regro/cf-scripts; the relock [action](https://github.com/regro/cf-scripts/actions/workflows/relock.yml) (which is responsible for updating the pinning pin) has been failing for reasons I haven't analysed yet; this is just to explain the mechanism

tkelman:
Not 100% sure but it looks like the relock failures are due to an incompatibility between `actions/checkout@v6` and `peter-evans/create-pull-request` prior to 7.0.9, tagging and propagating this https://github.com/conda-incubator/relock-conda/pull/441#issuecomment-3623086849 might resolve it

PR #8122, opened 2026-01-11
Update list of packages with free threading support

zklaus:
This PR brings an update of the list of packages with published upstream support for Python 3.14t, that is with wheels published on PyPI for that ABI.

It was created using https://github.com/zklaus/simple-index-db.

Pinging @h-vetinari, @isuruf.


jaimergp:
I guess this comes from the `show-free-threaded` pixi task, which seems to be supported by [this command](https://github.com/zklaus/simple-index-db/blob/fa8dfa80e9dc490c08e8fd7f4993721a9df2873b/src/simple_index_db/main.py#L283), which relies on this [DB query](https://github.com/zklaus/simple-index-db/blob/fa8dfa80e9dc490c08e8fd7f4993721a9df2873b/src/simple_index_db/main.py#L216-L237) (which I don't know how it really works). The PyPI<>conda mapping is [supported](https://github.com/zklaus/simple-index-db/blob/fa8dfa80e9dc490c08e8fd7f4993721a9df2873b/src/simple_index_db/conda.py#L5) by parselmouth, I see. When I [tried](https://github.com/conda-incubator/conda-pypi/blob/75ba850a466dc6dcf3ba949789e8d8e2d8cf534a/conda_pypi/dependencies/__init__.py) mappings in `conda-pypi` I needed all three mappings to work in tandem to get the most results out of it. Maybe you find some more packages with that logic (or a similar one).

I guess the list is ok, but I'll wait for others who have been more involved with the free-threaded efforts to chime in.

PR #8293, opened 2026-02-28
[arm_variant_type migration] Only expand arm_variant_type if defined

traversaro:
Without this fix, defining touch_arm_variant_type as suggested results in a re-rendering failure.

fyi @conda-forge/cuda

Checklist


recipe/migrations/arm_variant_type.yaml
```diff
@@ -52,7 +52,7 @@ __migrator:
     ```yaml
     context:
       # ensure arm_variant_type gets detected as a used variable
-      touch_arm_variant_type: ${{ arm_variant_type }}
+      touch_arm_variant_type: ${{ arm_variant_type if (linux and aarch64 and cuda_compiler_version != "None") else "None" }}
```
h-vetinari: There's a cleaner way in v1 recipes that I hadn't been aware of until very recently, which might be preferable

```yaml
build:
  variant:
    # ensure arm_variant_type gets detected as a used variable
    use_keys:
      - arm_variant_type
```

or, with the new, stricter rattler-build, probably also behind a `if: linux and aarch64 and cuda_compiler_version != "None"` guard.

h-vetinari (commented):
There's some more specific syntax than the (IMO pretty hacky) `touch_` resp. fake selector. Not as a criticism of this PR (which is fine IMO), but as a potential improvement.

carterbox:
I've synthesized an update to the migrator text based on the discussion here. I have tested that the new suggestion works in the torchcodec-feedstock.

PR #8395, opened 2026-04-02
Pin SUNDIALS to 7.5 and add migrator to 7.7

Saransh-cpp:
Checklist

XRef #8392 

Pin SUNDIALS to `v7.5` and add migrator for `v7.6`.

I am not sure what the right thing to do here is, given that several repositories use very different versions of SUNDIALS, such as:
- https://github.com/conda-forge/fmipp-feedstock/blob/main/recipe/meta.yaml: `2.7.*`
- https://github.com/conda-forge/pycvodes-feedstock/blob/main/recipe/meta.yaml: `>=5.3,<6.0`
- https://github.com/conda-forge/sunode-feedstock/blob/main/recipe/meta.yaml: `>=5.5.0,<6`
- https://github.com/conda-forge/muq-feedstock/blob/main/recipe/meta.yaml: `<7`

cc: @traversaro


h-vetinari (approved):
Thanks!

There's already 7.7 (https://github.com/conda-forge/sundials-feedstock/pull/131), I guess we could migrate for that right away?

h-vetinari:
> I am not sure what the right thing to do here is, given that several repositories use very different versions of SUNDIALS, such as:
>   * https://github.com/conda-forge/fmipp-feedstock/blob/main/recipe/meta.yaml: `2.7.*`
>   * https://github.com/conda-forge/pycvodes-feedstock/blob/main/recipe/meta.yaml: `>=5.3,<6.0`
>   * https://github.com/conda-forge/sunode-feedstock/blob/main/recipe/meta.yaml: `>=5.5.0,<6`
>   * https://github.com/conda-forge/muq-feedstock/blob/main/recipe/meta.yaml: `<7`

It depends how those feedstocks use sundials; if they're compiling against `libsundials_*`, then they are exposed to the ABI and need to respect the runtime version-constraint imposed by the run-export. That means to get a consistent set of packages across conda-forge, those caps should be removed (where possible), which is a prerequisite for the bot to populate the variant config for sundials with the version from the pinning that's being added here

Saransh-cpp:
Thanks, @h-vetinari! I have updated the migrator to migrate to 7.7.

h-vetinari (approved):

PR #8555, opened 2026-06-01
Drop dual HDF5 migration

hmaarrfk:
I feel like it has been long enough???

what do you all think?

cc: @conda-forge/hdf5 @h-vetinari 

Checklist

### DIFF
--- a/recipe/migrations/hdf52.yaml
+++ b/recipe/migrations/hdf52.yaml
@@ -4,9 +4,10 @@ __migrator:
   kind: version
   migration_number: 1
 # HDF5 likely changed alot of its API
-# During the initial phase of the migration lets build for both
-# HDF5 2 and the latest version of HDF5
+# The initial transition phase built for both HDF5 2 and 1.14.6.
+# We now only migrate remaining feedstocks to HDF5 2.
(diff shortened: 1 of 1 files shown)


h-vetinari (commented):
Hm. I wouldn't be in favour of closing the [migration](https://conda-forge.org/status/migration/?name=hdf52), since we're barely at 50% merged PRs. Just dropping 1.x from the migrator seems reasonable, though it does imply that we think we can finish the rest in 2-3 months. Do we think that's feasible?

https://github.com/conda-forge/vtk-feedstock/pull/436 has a lot of children in the migration, and will need a new vtk version, which is apparently expected mid-summer according to @vicentebolea. If we can get vtk out of the way in that 2-3 months timeframe, then I think we'll be able to make a sufficient dent in the remaining PRs to close the migration by latest ~August.

hmaarrfk:
The main motivation for me is to break the opencv matrix. 

Qt and not qt. 
hdf5 2 and 1

4 builds per Python is a lot

perhaps I can find an opencv specific solution

hmaarrfk:
https://github.com/conda-forge/opencv-feedstock/pull/518 so with this,

PR #8732, opened 2026-07-16
Fix migration test set insertion

ZupoLlask:
## Summary

- Fix a typo in the local fallback path of `test_timestamps_against_main`.
- Use the set's `add` method so direct local migration tests can collect new files.

## Validation

- `pytest -vvs recipe/test_migrations.py`: 96 passed.
- `conda-smithy recipe-lint recipe`: passed.
- Targeted pre-commit checks: passed.
- `git diff --check`: passed.

This is a separate test-only contribution with no recipe, migration, or generated CI changes.


h-vetinari (approved):
Thanks for catching this; baffling that this didn't already blow up upon introduction

PR #8763, opened 2026-07-23
Regenerate free-threaded-314.txt with simple-index-db

killua156:
Regenerated with `simple-index-db` (+202, no removals).


isuruf:
Can you not remove any packages here?

killua156:
done isuru. Will try to make it use additive/union logic rather than delta.

isuruf:
Thanks for the PR

recipe/migration_support/free-threaded-314.txt
```diff
@@ -1,11 +1,11 @@
-# This file was created with simple-index-db 0.2.0.
```
h-vetinari, line 1: You should not have "sorted" the comment at the top of the file.

h-vetinari: Fixed in 99d90a03476e9b8cb9d324eb5c1640fbc2a2b3bf

PR #8764, opened 2026-07-23
NEW: Migrate to dav1d>=1.3.0

carterbox:
Checklist

Create a mini-migrator which will migrate `dav1d` to `dav1d-devel`. Starting with `dav1d=1.3.0`, the `dav1d` package was split into runtime and development components and allowed looser run_exports. Starting with `dav1d=1.5.4`, the development package uses `-devel` instead of `-dev`. The current channel pinnings are `dav1d=1.2.1`. The latest `dav1d` version is `1.5.4`.

Depends on https://github.com/conda-forge/conda-forge-bot/pull/6356

cc: @conda-forge/dav1d

### DIFF
--- a/recipe/migrations/dav1d130.yaml
+++ b/recipe/migrations/dav1d130.yaml
@@ -1,0 +1,8 @@
+__migrator:
+  build_number: 1
+  commit_message: Rebuild for dav1d 1.3.0 and rename dav1d to dav1d-dev
+  kind: version
+  migration_number: 1
+dav1d:
+- '1.3.0'
+migrator_ts: 1784845364
--- a/recipe/migrations/dav1d130.yaml
+++ b/recipe/migrations/dav1d130.yaml
@@ -1,8 +1,8 @@
 __migrator:
   build_number: 1
-  commit_message: Rebuild for dav1d 1.3.0 and rename dav1d to dav1d-devel
+  commit_message: Rebuild for dav1d 1.5.4 and rename dav1d to dav1d-devel
   kind: version
   migration_number: 1
 dav1d:
-- '1.3.0'
+- '1.5.4'
 migrator_ts: 1784845364
--- a/recipe/migrations/dav1d130.yaml
+++ b/recipe/migrations/dav1d130.yaml
@@ -1,6 +1,6 @@
 __migrator:
   build_number: 1
-  commit_message: Rebuild for dav1d 1.3.0 and rename dav1d to dav1d-dev
+  commit_message: Rebuild for dav1d 1.3.0 and rename dav1d to dav1d-devel
   kind: version
   migration_number: 1
 dav1d:


h-vetinari:
Can we please stay with `-devel` as a suffix (at least for non-nvidia packages)?

carterbox:
This feedstock has been using the -dev suffix since Feb 2024, but because of the channel pinnings no one has been using it in our recipes.

What is the standard method of migrating non-feedstock consumers of a package?

recipe/migrations/dav1d130.yaml
```diff
@@ -0,0 +1,8 @@
+__migrator:
+  build_number: 1
+  commit_message: Rebuild for dav1d 1.3.0 and rename dav1d to dav1d-dev
```
carterbox:
```suggestion
  commit_message: Rebuild for dav1d 1.3.0 and rename dav1d to dav1d-devel
```

h-vetinari:
> This feedstock has been using the -dev suffix since Feb 2024,

Hm. I guess we could also grandfather this in, though I'm not excited about further propagating this divergence.

> What is the standard method of migrating non-feedstock consumers of a package?

We have very little idea about what non-feedstock consumers do. The standard way would be to add
```yaml
outputs:
    # actual development package
  - name: dav1d-devel
    [...]
    requirements:
      [...]
      run_constrained:
        # avoid co-installation with old version of previous name of devel package
        - dav1d-dev =={{ version }}

  # compatibility output
  - name: dav1d-dev
    build:
      run_exports:
        # same as for -devel
    requirements:
      host:
        - {{ pin_subpackage("dav1d-devel", exact=True) }}
      run:
        - {{ pin_subpackage("dav1d-devel", exact=True) }}
    test:
      commands:
        - echo "pure metapackage"
```

After some time (a year?), when you can be reasonably sure that consumers have adapted to the new world, you can remove the compatibility outputs.

hmaarrfk:
hmm, you know how to realize the bugs immediately after you press merge

we should add 

```
dav1d-devel
```
to the migrator pinnings too.

hmaarrfk:
@carterbox do you want to make that followup PR?

h-vetinari:
> do you want to make that followup PR?

That responsibility falls to the person clicking the merge button, IMO. Either you pause the migration, or you fix it yourself. But since all of this affects live infrastructure and the bots will simply run with whatever they get (and cause extra work to fix afterwards), we shouldn't expect someone to respond ASAP on the weekend.

recipe/migrations/dav1d130.yaml
```diff
@@ -0,0 +1,8 @@
+__migrator:
+  build_number: 1
+  commit_message: Rebuild for dav1d 1.5.4 and rename dav1d to dav1d-devel
+  kind: version
+  migration_number: 1
+dav1d:
+- '1.5.4'
```
h-vetinari, line 7: next time when updating the version, please rename the migrator file appropriately too

h-vetinari: I missed that the filename was linked to the minimigrator in the bot - I'm sorry about that. Fix is in #8906

carterbox:
We don't need the pinnings after migrating to dav1d-devel because the run_exports in newer versions are less strict.

h-vetinari:
> We don't need the pinnings after migrating to dav1d-devel because the run_exports in newer versions are less strict.

Is dav1d's ABI guaranteed to be eternal? I see there's not even a `<2` in the run-export...

PR #8817, opened 2026-08-05
Revert "skip special migrations in test_timestamps_against_main"

h-vetinari:
For https://github.com/conda-forge/conda-forge-bot/pull/6417, as diagnosed by @beckermr. I remember adding this after debugging some issue in #7777, but since this change ultimately got broken out into another PR (#7900), I can't really reconstruct what that was. I guess we'll see here in CI


h-vetinari:
Well, seems like this passes, so I guess we can just merge? Will give it some time for comments though

PR #8822, opened 2026-08-06
Add clang and clangxx to linux-riscv64 migrator

luhenry:


luhenry:
cc @isuruf @h-vetinari

luhenry:
There seem to be some [GitHub Actions incident](https://www.githubstatus.com/incidents/qcvjkzcs7j74)

recipe/migration_support/linux_riscv64.txt
```diff
@@ -1,4 +1,7 @@
 anaconda-client
+clang
+clangxx
```
h-vetinari, line 3: This should have used feedstock names (i.e. clangdev, compiler-rt), instead of package names. The conda-forge page has a dedicated [search function](https://conda-forge.org/packages/) for finding the feedstock for a given package.

luhenry: I didn’t know of this search function and it’s a lot easier than guessing the feedstock for the package name. Thanks!

PR #8849, opened 2026-08-11
close pybind11_abi migration

h-vetinari:
This migration has been open for over a year, and not much of anything is happening anymore. Also, there's the next migration waiting in the wings already (#8838). So I suggest to just close this one.

CC @conda-forge/pybind11

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -970,7 +970,7 @@ pulseaudio_client:
 pulseaudio_daemon:
   - '17.0'
 pybind11_abi:
-  - 4
+  - 11
 pythia8:
   - '8.312'
 python:
--- a/recipe/migrations/pybind11_abi11.yaml
+++ b/recipe/migrations/pybind11_abi11.yaml
@@ -1,8 +1,0 @@
-__migrator:
-  build_number: 1
-  commit_message: Rebuild for pybind11 3 and pybind11_abi 11
-  kind: version
-  migration_number: 1
-migrator_ts: 1752454025.946169
-pybind11_abi:
-- '11'


traversaro:
I totally agree, I was bite by the pybind11-abi==12.* change recently, and at least wrapping up the pybind11-abi==11.* migration would be great.

carterbox:
As a new core member, I'm curious about how we judge if a migration has gone stale (stopped making progress). You say it has been open for a year, but shouldn't we judge by when the youngest PRs in the migration were opened? The youngest are only 3 weeks. Or the main motivation is that a new ABI of pybind is available?

h-vetinari:
> I'm curious about how we judge if a migration has gone stale (stopped making progress).

There's no hard and fast rules. The bot will consider it done once all PRs are opened (even if the majority of them are unmerged; see https://github.com/conda-forge/conda-forge.github.io/issues/2276). Assuming that even just one child remains (because of a stale/unmaintained parent), the bot will never close it though. There's various metrics one can use as an aid for making that decision: age of the migrator, number/percentage of feedstocks still to-do (and how maintained they are, resp. whether they're leaf packages, or have many children), or whether a new version to migrate has already appeared.

For me personally, usually before closing larger migrations (boost, pytorch, arrow, abseil, etc.), I go through the open PRs on the migrator page and try to fix what I can (+ping the maintainers), noting which feedstocks could be fixed in principle, and which ones are just dead (or at least comatose); arguably the latter ones could be archived, but that was always a trigger for wider discussions, so usually I just move on; xref https://github.com/conda-forge/conda-forge.github.io/issues/1724.

Examples:
* https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/8616
* https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/8577
* https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/8383

But ultimately, there's a pretty wide spectrum of how "finished" a migration is when we close it. Certainly, it'd be great if more people keep an eye on stuck migrations and manually try to unblock individual feedstocks (whether member of cf/core or not! as non-core one can just open a new PR based on the bot branch).

I don't consider myself too deeply involved in pybind, so I didn't want to make that call for the maintainers, but FWICT it seems to me like the conditions are there to close it and then migrate the next version.

henryiii:
I'd be fine with that. :)

h-vetinari:
OK, let's do this. @traversaro et al., please feel free to open a PR with a migrator for `pybind11_abi` v12.

PR #8860, opened 2026-08-11
Reinstate stdlib pins and absl migrator

h-vetinari:
Fix for #8853


h-vetinari:
Test failure is spurious, in the sense that we need to restore the migration with the same timestamp

PR #8891, opened 2026-08-25
add paused py3.15 migration; close py3.14 migration; drop py3.10

h-vetinari:
https://github.com/conda-forge/python-feedstock/pull/900 has been merged, so we should be ready to migrate here.

Xref #7598

I'm not proposing to close the 3.14t migration, not all feedstocks are ready for this. We should probably do as we did for 3.13t vs. 3.14t, i.e. remigrate for 3.15t (c.f. #7682), and then eventually drop the 3.14t migrator (c.f. #8262)

CC @conda-forge/core @conda-forge/python

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -974,17 +974,16 @@ pythia8:
 python:
   # win-arm64 on conda-forge supports only 3.14+
   # part of a zip_keys: python, is_python_min
-  - 3.10.* *_cpython   # [not ((win and arm64) or riscv64)]
   - 3.11.* *_cpython   # [not ((win and arm64) or riscv64)]
   - 3.12.* *_cpython   # [not ((win and arm64) or riscv64)]
   - 3.13.* *_cp313     # [not ((win and arm64) or riscv64)]
-  - 3.14.* *_cp314     # [(win and arm64) or riscv64]
+  - 3.14.* *_cp314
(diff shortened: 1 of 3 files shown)


isuruf (changes requested):
You need to set channel_sources to use the pre-release labels

h-vetinari:
Good point, thanks for catching that.

h-vetinari:
Pre-commit keeps failing with "timeout waiting for merge ref"; I haven't found a way to restart it (without closing/reopening the PR, which is very noisy)

h-vetinari:
In the [announcement](https://conda-forge.org/news/2026/08/31/python-3-10) from https://github.com/conda-forge/conda-forge.github.io/pull/2923 that was merged a couple hours ago, we said (copying the same wording we had for 3.14)
> we have decided to drop 3.10 from our default build matrix. This will be reflected in your feedstock configuration on the next rerender.

Let's make this actually match reality. I've looked at doing this as a separate PR, but due to the pinning&zip changes, it'd be quite a bit messier to do this separately. However, I've paused the 3.15 migration (https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/8891/commits/98fd97c65e6fc1bf6500e1546f0df9d2df930307) while we get everything ready (e.g. https://github.com/conda-forge/conda-forge.github.io/pull/2925).

PR #8902, opened 2026-08-29
Pin Flang 22 on Windows ARM64

ClayWarren:
Enables the published Flang 22 compiler for the Windows ARM64 target in global pinning.


h-vetinari (approved):
thanks. I definitely want to wrap up the overall move to LLVM-based flang as the default fortran compiler on windows at some point 🥲

h-vetinari:
> Verified with a fresh win-arm64 solve selecting flang_win-arm64 22.1.8 build 1.

I don't know how your clanker thinks it verified this, but that's unlikely. The only way to do that is by copying this file somewhere, and then rerendering a relevant feedstock (e.g. your openmp PR) with `conda-smithy rerender -e ../path/to/this/state/of/global/pinning/conda_build_config.yaml`.

ClayWarren:
You're right. That line only verified package availability, not this pinning change. I removed it; I should have tested a consuming rerender with this PR's CBC as the exclusive config.

PR #8903, opened 2026-08-30
Dav1d migration follow-up

h-vetinari:
Follow-up to #8764. Due to the change of migrator name, this effectively restarts the migration; however, since the only so-far opened PRs
https://github.com/conda-forge/libavif-feedstock/pull/61 & https://github.com/conda-forge/ffmpeg-feedstock/pull/396 haven't been merged yet, this is not a big deal.


hmaarrfk:
Thanks.

recipe/conda_build_config.yaml
```diff
@@ -332,6 +332,8 @@ dartsim_cpp:
   - '6.19'
 dav1d:
   - 1.2.1
+dav1d_devel:
+  - 1.2.1
```
carterbox, line 336: This package doesn't exist, so I'm not sure why pinning it is necessary.

h-vetinari: It's to help the bot pick up the fact that upon closing the migration, it needs to update this pin.

PR #8906, opened 2026-08-30
use new `replacements:` feature for dav1d

h-vetinari:
Another follow-up to #8764, because in #8903 I missed (and thus broke) the relation of the filename with the minimigrator in the bot repo. However, this provides an opportunity to try out brand-new [infrastructure](https://github.com/conda-forge/conda-forge-bot/pull/6593) that creates these minimigrators on the fly, without having to patch the bot repo each time.


hmaarrfk:
Thanks

PR #8910, opened 2026-08-31
Pin unibilium to 2.1.4

thekpaul:
Selects the release that corrects unibilium's malformed shared-library SONAME for future conda-forge builds. Rebuilt dependents then receive the matching runtime requirement through unibilium run exports.

Checklist

Based on related issue from https://github.com/conda-forge/nvim-feedstock/pull/56.

Pins unibilium to 2.1.4, which corrects its malformed shared-library SONAME.
This makes future dependent rebuilds select the corrected ABI; unibilium run exports then provide the corresponding runtime constraint.
A companion repodata-patches PR constrains historical Nvim artifacts below 2.1.4, preventing their malformed `DT_NEEDED` entry from resolving with the corrected library.
No broad rebuild migration is included here.


recipe/conda_build_config.yaml
```diff
@@ -1135,6 +1135,10 @@ ucx:
   - '1.22'
 uhd:
   - 4.10.0
+unibilium:
+  # unibilium 2.1.4 corrects the malformed shared-library SONAME.
+  # https://github.com/neovim/unibilium/commit/e2a20519ae247f58cbf2ab763bb8f59ef2d6b2b9
```
danielnachun:
```suggestion
```
This comment will become stale and irrelevant when the next version of unibilium is released, let's remove it.

danielnachun (changes requested):

recipe/conda_build_config.yaml
thekpaul: Done!

PR #8562, opened 2026-06-02
Close out CUDA 13.0 Migration

carterbox:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #8926, opened 2026-09-09
CUDA 13.4 migration / Pinning bump

carterbox:
Checklist

Depends on #8562

Bumps the CUDA pinnings from 13.0 to 13.4 immediately.

A migration which traverses the graph in order is not strictly necessary because the run_exports of the CUDA compiler allow for packages migrating to 13.4 to coexist with older ones. There are no new architectures to target with CUDA 13.4. There are bug fixes and optimizations which each feedstock may adopt as is appropriate for them without holding other packages back.

Alt: Traditional Migration

~~Alt: Explicit migration, but tell the bot it can be unordered (opens all migration PRs at once). Would allow us to track progress of the migration.~~ I don't think this option exists for version migrations.


isuruf (commented):
Are there any architectures added by 13.4 that we have to rebuild all the downsteam packages for?

carterbox:
xref: https://developer.nvidia.com/cuda-13-4-0-download-archive

carterbox:
13.4 adds support for Rubin (sm107), but I believe this is for developer preview not for production.

13.4 adds support for GCC 16 and Clang 22 as host-side compilers with NVCC.

carterbox:
In the community meeting, Isuru suggested that a migration is not necessary because there are no new archs to target. Instead we can just bump the pinnings to 13.4 after the 13.0 migration closes.

h-vetinari:
> In the community meeting, Isuru suggested that a migration is not necessary because there are no new archs to target. Instead we can just bump the pinnings to 13.4 after the 13.0 migration closes.

It'll still break a bunch of build scripts that do stuff like
```bash
  if [[ ${cuda_compiler_version} == 12.9 ]]; then
    export TORCH_CUDA_ARCH_LIST="5.0;6.0;7.0;7.5;8.0;8.6;8.9;9.0;10.0;12.0+PTX"
  elif [[ ${cuda_compiler_version} == 13.0 ]]; then
    export TORCH_CUDA_ARCH_LIST="7.5;8.0;8.6;8.9;9.0;10.0;11.0;12.0+PTX"
  else
    echo "unsupported cuda version. edit build.sh"
    exit 1
  fi
```

On top of that, there's the question whether all our constraints everywhere are of the kind `>=13.0,<14`, rather than unintentionally pinning the minor version, also for stuff like
```yaml
requirements:
  build:
    - {{ stdlib("c") }}
    - {{ compiler("c") }}
    - {{ compiler("cuda") }}
    - some_package_depending_on_cuda    # [build_platform != target_platform]
```
Overall, I'd prefer migrating. Or we should come up with a minimigrator that switches stuff like the `TORCH_CUDA_ARCH_LIST` example above to the new setup or relying on env vars that are set by the respective activation or cuda/pytorch

carterbox:
Any feedstock that is currently built with CUDA 13.0 can be built with CUDA 13.4 because CUDA compatibility.

I can manually propose changes to the [6 feedstocks](https://github.com/search?q=org%3Aconda-forge+%22+%3D%3D+13.0%22&type=code) that are hardcoding `${cuda_compiler_version} == 13.0`.

> the question whether all our constraints everywhere are of the kind >=13.0,<14

Can you provide a specific example of this concern? The [run_exports for `cuda-nvcc=13.0` are `>=13.0,<14`](https://github.com/conda-forge/cuda-nvcc-feedstock/blob/7c3faf28bd4e236249e4d195291a3e9afc7b5cf3/recipe/meta.yaml#L64-L65), so we don't need to migration in topographic order because all packages should be already floating up to `cuda-version=13.4` unless the package is pinning cuda-version unexpectedly/unnecessarily.

carterbox:
I have examples of feedstocks with incorrect pinnings from the stragglers in the CUDA 13.0 migration. e.g. [the `gromacs-feedstock` which is manually pinning the `cuda-toolkit` metapackage](https://github.com/conda-forge/gromacs-feedstock/blob/717ec6d4a0060e5520397fe07f2a8cbdf52ed38c/recipe/meta.yaml#L47). However, we cannot worry about those feedstocks which missed the 13.0 migration.

h-vetinari:
> > the question whether all our constraints everywhere are of the kind >=13.0,<14
> 
> Can you provide a specific example of this concern? The [run_exports for `cuda-nvcc=13.0` are `>=13.0,<14`](https://github.com/conda-forge/cuda-nvcc-feedstock/blob/7c3faf28bd4e236249e4d195291a3e9afc7b5cf3/recipe/meta.yaml#L64-L65), so we don't need to migration in topographic order because all packages should be already floating up to `cuda-version=13.4` unless the package is pinning cuda-version unexpectedly/unnecessarily.

I understand that if all component pins are correct everywhere, things will work as intended. My point is that a single stray too-tight constraint somewhere can mess this up in a way that's not obvious. I'm not claiming such a case exists necessarily. If you're willing to chase down any eventual fallout, then that addresses my concern.

> I can manually propose changes to the [6 feedstocks](https://github.com/search?q=org%3Aconda-forge+%22+%3D%3D+13.0%22&type=code) that are hardcoding `${cuda_compiler_version} == 13.0`.

Fine by me. Thanks

